PR #379, opened 2026-08-22
Clean OU-III publication dependencies and legacy-law studies

mgrouch:
## Summary
- keep adaptation (including the measurement-only vertical-acceleration proxy), Live initialization, and magnetic handling before all stability theory
- remove the `Filtering and virtual-reference precedent` subsection from the rendered OU--III manuscript
- keep the OU--III analytical sensitivity study as five directions: tau-only, sigma_aw-only, r_S-only, sigma_aw+r_S coupled, and tau+r_S coupled
- make the two coupled OU--III sweeps follow the deployed `SpectralMSE` scaling (`c^(6/7)` for the sigma coupling and `c^(24/7) sqrt(T_S(tau)/T_S(c tau))`, reducing to `c^(41/14)` away from cadence clamps, for the tau coupling), never the retired cubic law
- keep the OU--II study as a `PhysicalMSE` coefficient sweep only; no Empirical-law arm is run
- keep historical one-way transition evidence only in archived results for provenance; the publication contains only the bidirectional deployed-SpectralMSE low–high–low transition study
- retain the low-motion OU--III degradation result without publishing the archived one-way rapid-transition table or figure
- fix the late tuner-chart publication contract and malformed accented matrix notation that caused CI failures

## Evidence boundary
Historical robustness and one-way transition artifacts may remain under `reports/results` for reproducibility/provenance. They are not manuscript evidence. The OU--III publication tree contains the current SpectralMSE sensitivity evidence and the bidirectional low–high–low transition evidence only.

No estimator behavior is changed by the publication cleanup.


mgrouch:
Merging it. But it failed to generate pdf of main article

PR #431, opened 2026-08-29
OU-III: add next-generation widened P4 certificate

mgrouch:
## Summary

Adds a next-generation P4 certificate that preserves the existing exact Cayley/information-metric theorem but tightens the nonlinear complete-word defect budget.

The legacy P4 charges every possible state operation in every IMU sample with one global worst-case defect constant. The new producer separates prediction, S=0, accelerometer, and magnetometer defect constants, exploits the exact linear S=0 residual, and transports the operation-class sum through the same P3 unit prefix-information bound. Rejected/not-due correction branches remain covered because charging each correction once per sample is still an upper bound.

The refinement is source-only: no replay, sampled trajectories, filter changes, weaker P3 margin, metric changes, or branch exclusions. It fails closed unless the new transported defect bound is no larger than legacy P4 and the certified W level is no smaller for both H and A. The legacy values are emitted alongside the widened values, including B reduction and W/sqrt(W) widening factors.

New tests require strict widening in both H and A, schema-4 enclosure compatibility, Cayley/quaternion branch safety, and continued A-mode bias-projection interior safety.

## Goal

Widen P4 outward independently of P5, then compare the larger inner funnel against the current P5 startup/outer enclosure so P5 only has to bridge any remaining gap.

## Summary by CodeRabbit

- **New Features**
  - Added next-generation P4 certificate generation with refined bounds for H and A modes.
  - Added full and compact JSON certificate output, including validation results and improvement metrics.
  - Added safety checks that prevent certificates from being emitted when refinement conditions are not satisfied.

- **Tests**
  - Added coverage for certificate validation, widened bounds, schema compliance, and mode-specific safety constraints.

- **Chores**
  - Added automated workflow support for building certificates and running the validation suite.

### DIFF
--- a/.github/workflows/ou3-p4-nextgen.yml
+++ b/.github/workflows/ou3-p4-nextgen.yml
@@ -5,6 +5,7 @@ on:
     paths:
       - "tools/ou3_p4_nextgen_widened_certificate.py"
       - "tools/ou3_p4_nextgen_directional_certificate.py"
+      - "tools/ou3_p4_nextgen_gain_certificate.py"
       - "tools/ou3_p4_nonlinear_word_certificate.py"
       - "tools/ou3_p4_group_algebra.py"
       - "tools/ou3_p4_exact_word_map.py"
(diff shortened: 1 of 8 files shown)


mgrouch:
This route was successful, not a dead end: it delivered the major P4 widening (about 4.74e6x legacy). Closing only because its successful refinements are now fully subsumed by consolidated PR #438, which continues from the stronger direct-contraction frontier.

PR #432, opened 2026-08-29
OU-III: widen P4 with radius continuation and exact prefix bootstrap

mgrouch:
## Summary

Adds an independent P4 widening route in parallel with #431. This branch starts from current `main` and does not depend on the operation-class / measurement-gain refinements in #431.

The new producer removes two conservative proof choices from legacy P4:

- it rebuilds the exact Cayley/quaternion nonlinear defect constant on a nested set of canonical proof balls instead of freezing the defect at one tiny design radius;
- it uses the actual prefix bootstrap `sqrt(W_prefix) <= sqrt(W0) [1 + B sqrt(W0)]` instead of replacing the bracket by the fixed factor 2 before solving the admissible level.

The endpoint still proves the same advertised contraction `W_end <= (1-delta/2) W0`, but uses the cancellation-free exact square-root budget `sqrt(1-delta)+B sqrt(W0) <= sqrt(1-delta/2)`.

This is source-only and theorem preserving: no replay, no sampled trajectories, no filter changes, no P3-margin changes, and no branch exclusions. Every candidate radius independently rebuilds outward-rounded defect bounds and must satisfy Cayley chart safety, deployed quaternion-series safety, the bootstrap condition, and A-mode bias-projection interior safety. The selected level must be monotone relative to legacy P4.

A focused CI workflow emits H/A legacy and radius-continuation W levels, widening factors, selected proof radius, prefix radius, and the active limiting constraint.

## Parallel intent

This PR is intentionally separate from #431 so both widening strategies can run concurrently. Once both produce numbers, the theorem-preserving refinements can be compared and, where independent, stacked in a follow-up integration.

### DIFF
--- a/.github/workflows/ou3-p4-radius-continuation.yml
+++ b/.github/workflows/ou3-p4-radius-continuation.yml
@@ -1,0 +1,63 @@
+name: ou3-p4-radius-continuation
+
+on:
+  pull_request:
+    paths:
+      - "tools/ou3_p4_radius_continuation_certificate.py"
+      - "tools/ou3_p4_nonlinear_word_certificate.py"
+      - "tools/ou3_p4_group_algebra.py"
+      - "tools/ou3_p4_exact_word_map.py"
+      - "tools/ou3_p4_node_metrics.py"
+      - "tools/ou3_explicit_information_word_certificate.py"
+      - "tools/ou3_proof_operating_domain.json"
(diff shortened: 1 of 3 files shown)


mgrouch:
Closing this experiment as a P4 dead end. Radius continuation alone passed, but when stacked with the stronger #431 refinements it produced no additional widening; the endpoint small-gain constraint remained active. The useful exact-prefix/radius machinery is retained in the consolidated #438 branch.

PR #434, opened 2026-08-29
OU-III: combine successful P4 widening routes

mgrouch:
## Summary

Both independent P4 widening experiments succeeded in CI, so this PR stacks their theorem-preserving refinements into one third-generation certificate instead of choosing between them.

From #431 it carries forward:
- operation-specific prediction/S/accelerometer/magnetometer nonlinear defect transport;
- directional measurement-operator bounds;
- cancellation-free exact endpoint budget;
- measurement-specific Kalman-gain envelopes.

From #432 it carries forward:
- nested proof-radius continuation that rebuilds the outward-rounded Cayley/quaternion defect constants at each proof radius;
- exact prefix bootstrap `sqrt(W_prefix) <= sqrt(W0) [1 + B sqrt(W0)]` and direct solution of the design/chart/projection caps.

The combined producer applies radius continuation *after* the #431 class-local gain/directional refinements, so each candidate radius recomputes class-specific nonlinear correction defects with the tighter gains. It fails closed unless the resulting H/A level is at least as large as the successful #431 final level, preserves schema-4 enclosure compatibility, remains in the deployed quaternion branch, and stays inside the A-mode projection interior.

## Prior CI results being stacked

#431: all focused P4 tests, `ou-validation`, and broad `build` passed. Final gain-refined P4 widened legacy W by about 4.7405e6x in both H and A.

#432: focused radius-continuation tests, `ou-validation`, and broad `build` passed. Radius continuation alone widened legacy W by 4x in both H and A.

This PR is based on current main (including #433) and has its own focused CI that emits legacy, #431-equivalent, and final combined W levels plus the selected proof radius and limiting cap.

## Summary by CodeRabbit

- **New Features**
  - Added next-generation certificate generation for combined H and A modes.
  - Added refined directional, gain, widening, projection, and safety validation.
  - Added JSON certificate output with pass/fail status, certified levels, widening factors, design metrics, and failure details.

- **Tests**
  - Added automated validation covering baseline compatibility, maximal radius continuation, safety bounds, schema compliance, and projection behavior.

- **Chores**
  - Added pull-request and manually triggered CI checks to compile, test, validate, and report certificate results.

PR #436, opened 2026-08-29
OU-III: widen P4 with support-aware information transport

mgrouch:
Runs the support-aware information-metric widening route in parallel with the exact-correction P4 work. It decomposes the nonlinear word budget so attitude-only prediction/S reset-composition defects are transported with an attitude-support information bound derived from the conditioned P3 covariance lower construction, while accelerometer/magnetometer full-state defect terms retain the global metric bound. The refinement is source-only, monotone, and fails closed if the conditioned covariance data are unavailable or the support bound does not dominate the global route. Focused CI emits H/A W before/after and the support-vs-global metric upper bounds. This PR is stacked on #434's branch so it can run now; after #434 merges it can be retargeted to main.

### DIFF
--- a/.github/workflows/ou3-p4-support-aware-metric.yml
+++ b/.github/workflows/ou3-p4-support-aware-metric.yml
@@ -1,0 +1,21 @@
+name: ou3-p4-support-aware-metric
+on:
+  pull_request:
+    paths:
+      - "tools/ou3_p4_support_aware_metric_certificate.py"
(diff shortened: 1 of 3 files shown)


mgrouch:
Closing this experiment as a P4 dead end. The support-aware metric refinement passed its focused certificate but widened W by only about 2.0e-6 relative, and its branch ou-validation was red. It does not justify carrying a separate proof route. The consolidated #438 branch keeps the successful exact-correction/direct-contraction line and continues widening there.

PR #437, opened 2026-08-29
OU-III: direct strict nonlinear P4 word contraction

mgrouch:
Runs the direct strict-return-map contraction route in parallel with #436 and the exact-correction back-port on #434. Instead of requiring the previous fixed endpoint decrease delta/2, it asks the minimal theorem question directly for each source-safe proof radius: sqrt(1-delta)+B(q)sqrt(W)<1. The positive gap 1-sqrt(1-delta) is evaluated cancellation-free and every candidate still closes prefix/design/Cayley/quaternion/A-projection safety. This is a first direct contraction layer over the validated exact-operation envelope; it is source-only and does not use replay. Focused CI emits H/A W before/after, widening factor, selected radius and limiting cap. It is stacked on #434 so it can run immediately and can be retargeted to main after #434 merges.

### DIFF
--- a/.github/workflows/ou3-p4-direct-word-contraction.yml
+++ b/.github/workflows/ou3-p4-direct-word-contraction.yml
@@ -1,0 +1,21 @@
+name: ou3-p4-direct-word-contraction
+on:
+  pull_request:
+    paths:
+      - "tools/ou3_p4_direct_word_contraction_certificate.py"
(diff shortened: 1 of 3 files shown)


mgrouch:
This route was successful and produced the strongest predecessor P4 level (~6.24328e-134, ~1.9e7x legacy). Closing only because #438 now contains this direct strict-contraction route and continues widening it with a self-consistent prefix bootstrap.

PR #453, opened 2026-09-01
Fix the two stale OU-III P5 validation expectations breaking main

mgrouch:
Fixes the three assertion failures in the main build's `ou-evidence / commit` job
([run 33541491462](https://github.com/bareboat-necessities/ocean-imu/actions/runs/33541491462/job/99979427013)).
They come from two independent causes; both are stale expectations in the checks, not
producer regressions.

## 1. Deployed prediction step compared against the wrong constant

`tools/ou3_p5_45deg_first_accel_tilt_yaw_source_bound.py`

`validate` checked the carried prediction step against the binary64 spelling of `0.005`
with a four-ULP window. The step it validates comes from `FREQ_SMOOTHER_DT`, which is
`constexpr float 1.0f / 200.0f` and is therefore parsed with the binary32 semantics the
source-domain contract applies everywhere else: `0.004999999888241291` s — about
`1.1e-10` s below the decimal value, roughly `1.3e8` binary64 ULPs outside the window.
The check could never pass, so the producer reported

```
next prediction step is not the deployed 5 ms interval
PASS carries validation failures
```

and `test_P1_tilt_intersection_is_only_deployed_startup_subroute` failed.

The validator now reads the constant back from the deployed header and compares against
that. The adjacent binary32 cadences are `4.7e-10` s away, so the few-ULP slack still
cannot hide a different deployed cadence, and the check is bound to the shipping constant
rather than to a decimal literal retyped in the validator.

## 2. V11/V12 coarse fixtures assumed a V10 prerequisite failure

`tests/validation/test_ou3_p5_sample1_structured_full_gain_v11.py`
`tests/validation/test_ou3_p5_sample1_structured_full_gain_v12.py`

Both semantic fixtures build on a deliberately coarse `4^3` grid and asserted that
`validate` returns exactly `["V10 prerequisite did not pass"]`. V10's own enclosure has
since tightened — the exact monotone gain enclosure lifted over the full source-cell-0
cover — and it now closes on that grid too: 64 joint cells, none unclosed, max combined
directional correction `8.2269` rad inside the 9 rad range. The parent no longer
contributes a prerequisite failure, so both tests failed on a stale expectation.

The tests now assert the fail-closed property the fixtures exist to pin: V11 leaves 31 of
64 joint cells unclosed and V12 all 64, so both certificates stay `NOT_ESTABLISHED` with a
recorded witness instead of inheriting V10's `PASS`, and `validate` reports no spurious
failure. The stale prose about V10 needing the focused `24^3` grid is corrected with it.

## Scope

No producer numerics, filter source, declared domain, or promoted claim changes.

## Verification

The three previously failing tests pass locally:

- `test_ou3_p5_45deg_sample1_h_group_norms` — 7 tests, OK
- `test_ou3_p5_sample1_structured_full_gain_v11` + `_v12` — 12 tests, OK

A full `unittest discover` was also run in this container; the remaining failures there are
environmental (no `numpy`/`matplotlib`, and the fingerprinted simulation records are not
fetched, so the archive-provenance and publication-tree checks cannot run) and are
unrelated to this diff. CI has those prerequisites.

https://claude.ai/code/session_01UF7yin7Q8UaFCraiTLtDGS

---
_Generated by [Claude Code](https://claude.ai/code/session_01UF7yin7Q8UaFCraiTLtDGS)_

### DIFF
--- a/tests/validation/test_ou3_p5_sample1_structured_full_gain_v11.py
+++ b/tests/validation/test_ou3_p5_sample1_structured_full_gain_v11.py
@@ -13,14 +13,22 @@ class StructuredFullGainV11Tests(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         # This deliberately coarse fixture is a semantic/fail-closed test, not
-        # the authoritative V10/V11 numerical certificate. V10 requires the
-        # focused 24^3 grid to close; the workflow emitter below supplies that.
+        # the authoritative V11 numerical certificate; the workflow emitter
+        # supplies the focused 24^3 grid for that.
         cls.d=V11.build(source_pieces=4,source_cell_index=0,p_pieces=4,tangent_pieces=4,axial_pieces=4)
 
-    def test_validate_fail_closed_coarse_prerequisite(self):
+    def test_validate_fail_closed_on_the_coarse_grid(self):
+        # V10's own enclosure has since tightened enough to close on this
+        # coarse grid, so the parent no longer contributes a prerequisite
+        # failure here. What the fixture still pins is the fail-closed rule
+        # that matters: V11's own joint cells do not all close on 4^3, so the
+        # certificate stays NOT_ESTABLISHED and validate() reports the witness
+        # rather than a spurious failure or an inherited PASS.
         failures=V11.validate(self.d)
-        self.assertEqual(failures,["V10 prerequisite did not pass"])
+        self.assertEqual(failures,[])
         self.assertEqual(self.d["P5_SAMPLE1_PSD_S_PERTURBATION_V11"],"NOT_ESTABLISHED")
+        self.assertGreater(self.d["unclosed_joint_cells"],0)
+        self.assertIsNotNone(self.d["first_unclosed_joint_cell"])
 
     def test_omitted_covariance_terms_are_now_explicit(self):
         for k in (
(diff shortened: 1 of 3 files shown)


mgrouch:
The red `ou-validation / validate` check on this PR is not this PR's failure, and it is not something this PR can turn green. It fails before any test runs:

```
python3 ../../tools/ou_evidence_contract.py --auto
validation: replay dependency differs from replay provenance: src/kalman_ou_ii/SeaStateFusionFilter_OU_II.h
validation: replay dependency differs from replay provenance: tests/kalman_ou_ii/kalman_ou_ii-sim.cpp
make: *** [Makefile:158: evidence-contract] Error 1
```

That is the state of `main` itself. Running the same contract on a clean checkout:

- `ba33149` (before #452) — `OU evidence contract is current`, exit 0
- `36aa126` (current `main`, the merge of #452) — the two lines above, exit 1

#452 armed the OU-II front-end vibration guard, which changes `SeaStateFusionFilter_OU_II.h` and the OU-II sim, so the committed OU-II replay evidence no longer matches its provenance. The other open PRs are red on the same check for the same reason.

It is also downstream of exactly what this PR fixes. On `main`, `ou-validation` runs in `full` mode: regenerate → validate (`make test`) → commit the regenerated evidence. In [run 33541491462](https://github.com/bareboat-necessities/ocean-imu/actions/runs/33541491462/job/99979427013) the validate step failed on the three assertions fixed here, so *Commit the regenerated evidence* was skipped and the stale OU-II provenance stayed committed. Every PR's smoke `validate` has hit it since.

Regenerating that evidence is a full replay — the job `main` runs, roughly an hour of runners — so porting a fix into this PR is not possible here, and hand-editing provenance would defeat the contract. Once this merges and `main`'s next full run validates and commits the regenerated bundles, the PR smoke check goes green again.

The three assertions this PR does fix were verified locally against the current tree:

- `test_ou3_p5_45deg_sample1_h_group_norms` — 7 tests, OK
- `test_ou3_p5_sample1_structured_full_gain_v11` + `_v12` — 12 tests, OK

---

PR #464, opened 2026-09-02
Header ODR/self-containment fixes, plus a sanitizer, static-analysis and coverage layer

mgrouch:
Two changes, kept as separate commits.

1. **`0c4d684` + `605ffb6`** — the reported `NmeaChecksum.h` defects, plus the same defects found by auditing every header in `src/`.
2. **`700a71e`** — the missing sanitizer / static-analysis / coverage layer.

---

# 1. Header ODR and self-containment

Both reported defects reproduce exactly.

**Not self-contained** — `uint8_t` with no `<stdint.h>`:

```
error: 'uint8_t' does not name a type
note: 'uint8_t' is defined in header '<cstdint>'; did you forget to '#include <cstdint>'?
```

**ODR violation** — defined in a header without `inline`, so two TUs fail to link:

```
/usr/bin/ld: multiple definition of `nmea0183_checksum(char const*)'; a.o: first defined here
```

## Audit

Compiling every header in `src/` standalone, then linking it from two translation units:

| | link cleanly | fail standalone | duplicate definitions |
|---|---|---|---|
| before | 48 | 10 | 3 |
| after | 53 | 8 | 0 |

| header | defect |
|---|---|
| `nmea/NmeaChecksum.h` | `uint8_t` without `<stdint.h>`; non-inline definition |
| `nmea/NmeaXDR.h` | `gen_nmea0183_xdr` non-inline |
| `avg/KalmanSmoother.h` | `fabs` without `<cmath>`; three non-inline definitions |
| `discrete/MinMaxLemire.h` | `uint32_t` without `<stdint.h>`; non-inline definition |

`nmea/NmeaCompass.h` was already correct (`static inline` throughout). Forward declarations sitting immediately above the definition they preceded guarded nothing, so they are dropped rather than duplicated.

## Removing `-Wl,--allow-multiple-definition`

It tolerated exactly the duplicates above. No sketch needs it: each sketch is a single TU, and the only other TUs in the Arduino link (`ImuCalWizardRunner.cpp`, `W3dSimCommon.cpp`) include none of these headers. **All three `build-MCU` matrix jobs pass without it.**

---

# 2. Quality gates

There was no systematic sanitizer, static-analysis or coverage layer. Warning flags were partial: 5 of 15 test directories set `-Wall -Wextra`, 2 of those added `-Werror -pedantic`, and `CMakeLists.txt` set none.

`tools/quality_gates.sh` has one subcommand per gate; CI shells out to it, so a red gate reproduces locally with one command.

| gate | baseline when added | now |
|---|---|---|
| warnings | 1 real finding | 52 TUs clean under `-Werror` |
| sanitizers | never run | 6 suites, 0 ASan / 0 UBSan / 0 leak |
| cppcheck | 5 findings (3 false positives) | clean |
| clang-tidy | 18 findings, 16 noise | 2 blockers, below |
| python | 96 ruff findings / 122 files | clean |
| coverage | no tooling | 65.6% lines, floorless |

The codebase was in far better shape than the missing tooling suggested — the warnings gate cost exactly one real fix to turn on.

## What the gates caught

**`src/ahrs/KalmanQMEKF.h` — `-Wreorder`.** The `QuaternionMEKF` mem-init list runs `Q, Racc, Rmag, R`; the members declare `Racc, Rmag, R, Q`. Harmless today because no initializer reads another member, but the constructor already binds them in a deliberate order, so the mismatch is precisely the trap that bites whoever adds the first dependency.

**Python.** Eight dead imports, two placeholder-free f-strings, a function-call default argument, an unused loop variable, and a re-raise that swallowed its context. Two apparently-unused imports are kept with `# noqa` and a reason — both are imported for their side effects, and deleting them would have broken the module they register.

`tools/ou_validation.py` also closed over the loop variable `scenario` while binding `family` as a default argument *in the same signature*. Same latent-inconsistency class as the `-Wreorder`; benign at today's call site.

## Deliberately narrow

Every disabled check was disabled against a measured count on this tree, and documented where it is turned off. Two worth calling out:

- **`cert-msc32-c`** objects to the constant RNG seed. That seed is *required* — the simulations must replay bit-for-bit or the evidence contract breaks.
- **`black`** wants to reformat 120 of 122 Python files. A formatting diff that size buries real findings, so the Python gate checks defects only.

## The frozen replay closure

`tests/validation` pins the committed OU evidence to 34 immutable implementation files. Any byte change fails `validate`. **These gates never rewrite those files, and are built so an autofix cannot either** — `quality_gates.sh` has no `--fix` mode for that reason.

That constraint costs one piece of coverage. `clang-tidy` runs `continue-on-error` because:

```
src/kalman_ou_ii/SeaStateFusionFilter_OU_II.h:1375: error: constexpr variable
  'C_HS' must be initialized by a constant expression
```

`std::sqrt` in a `constexpr` initializer is a GCC extension; clang rejects it, and clang-tidy uses the clang frontend, so every TU including that header fails to parse. **This is also why the library does not currently build with clang at all.** The fix is two lines per file and bitwise identical (`0x3FB504F3`, verified), so it changes no numerical result:

```diff
-        constexpr float K = std::sqrt(2.0f) / std::numbers::pi_v<float>;
+        constexpr float K = std::numbers::sqrt2_v<float> / std::numbers::pi_v<float>;
```

Both files are frozen, so it needs an evidence regeneration to land — a maintainer decision, not a side effect of adding CI. `docs/quality-gates.md` records the patch, and dropping `continue-on-error` is a one-line follow-up.

## Scope choices

Sanitizers cover six unit-test suites, not all fourteen: the OU/TFG/AHRS/NLO sims run past 15 minutes each under ASan+UBSan (a first attempt timed out three of them at 900 s). `SAN_JOBS` defaults to 2 because instrumenting the Eigen instantiations OOM-killed `cc1plus` at `-j4` on a 16 GB machine. Coverage is a floorless report — 65.6% reflects six of fourteen directories, so setting a floor from it now would ratchet on the wrong baseline. All of this is in `docs/quality-gates.md`.

---

## Two header fixes deferred

`src/kalman_common/SeaStateFusionFilterCommon.h` (uses `Eigen::Vector3f` with no Eigen include) and `src/wave_dir/KalmanWaveDirection.h` (two non-inline test helpers behind an `#ifdef`) are both real defects and both in the frozen closure. Fixing them failed `validate`, so `605ffb6` reverts them for the same evidence-regeneration reason. Neither affects the audit numbers.

## Note on `validate`

The `validate` job is currently red **on `main` itself**, from a stale `PRODUCERS` list in `tests/validation/test_ou3_result_determinism.py`. Established by running the job's exact command against two clean worktrees — the failure sets are byte-identical. Detail and a proposed patch in [this comment](https://github.com/bareboat-necessities/ocean-imu/pull/464#issuecomment-5515806409).

## Verification

- Both original defects reproduced before, gone after.
- Two-TU link audit over all 61 headers: 0 duplicate definitions.
- Warnings gate verified to *fail* on a deliberately reintroduced `-Wreorder`, not just to pass.
- `make -C tests/validation evidence-contract` → `OU evidence contract is current`.
- Validation failure set identical to `main` before and after the Python edits.
- No file in the frozen closure is touched (checked against the manifests).

https://claude.ai/code/session_011iqRNvV3uMDrSk2yrmMJuj

## Summary by CodeRabbit

* **Improvements**
  * Improved build reliability by tightening symbol validation during compilation.
  * Added automated checks for warnings, sanitizers, static analysis, Python code quality, and test coverage.
  * Added a local command for running the same quality checks used in continuous integration.
  * Improved diagnostic messages when validation data is missing.
  * Corrected initialization ordering and header compatibility without changing runtime behavior.

* **Documentation**
  * Added guidance for running and interpreting project quality checks.


mgrouch:
## `validate` is failing, and it is not this PR

`validate` is red on head `605ffb6`, but the same failures reproduce on pristine `main` (`2492f5c`) with none of this PR's changes.

I ran the job's exact command — `make -C tests/validation test` — in one environment against two clean worktrees, `origin/main` and this PR's head. The failure sets are byte-identical:

```
$ diff <(main failures) <(PR head failures)
   (no output — identical)
```

Both exit 2. This PR touches four headers and one workflow file; it contains no Python and nothing under `tools/` or `tests/validation/`.

For the record, the earlier `validate` failure on `0c4d684` *was* mine — the evidence-contract error from touching two replay-provenance headers. That is fixed in `605ffb6`, and `make -C tests/validation evidence-contract` now reports `OU evidence contract is current`. What remains is a separate, pre-existing failure.

### Root cause

`tests/validation/test_ou3_result_determinism.py` hardcodes a `PRODUCERS` list, and three of its four entries were deleted from `tools/` by commits already on `main`:

| missing producer | deleted by |
|---|---|
| `ou3_p4_augmented_complete_word_design_v5.py` | `234cc7b` Remove superseded long-prefix P4 producer |
| `ou3_p4_terminal_cluster_p2_reduction.py` | `bd6eabe` Remove obsolete fixed terminal-cluster P4 reduction |
| `ou3_p4_terminal_source_equivalence.py` | `b22c9f2` Remove obsolete terminal-cluster P4 quotient |

The list was not updated alongside the deletions, so the test asserts on paths that no longer exist:

```
AssertionError: False is not true : current producer missing:
  .../tools/ou3_p4_augmented_complete_word_design_v5.py
```

`tests/validation` also fails `test_no_superseded_p4_fixed_schedule_routes_remain`, `test_publication_tree_has_no_legacy_numerical_law_studies`, and `test_changed_estimator_cannot_be_laundered_by_restat_then_auto_in_archive` on `main`. Those are contracts over the in-flight P4 cleanup rather than a stale path list, so I have not proposed a patch for them.

### Proposed patch (not applied here)

For the producer list specifically, the fix is to drop the three deleted entries, leaving the one producer that still exists:

```diff
--- a/tests/validation/test_ou3_result_determinism.py
+++ b/tests/validation/test_ou3_result_determinism.py
@@
 PRODUCERS = [
     ROOT / "tools" / "ou3_p2_clock_phase_tuner_graph.py",
-    ROOT / "tools" / "ou3_p4_augmented_complete_word_design_v5.py",
-    ROOT / "tools" / "ou3_p4_terminal_cluster_p2_reduction.py",
-    ROOT / "tools" / "ou3_p4_terminal_source_equivalence.py",
 ]
```

I have deliberately not applied it. Whether those three producers are meant to be gone for good or to come back under new names is a question about the P4 route cleanup — plausibly overlapping #463 — and answering it by editing a determinism contract from inside a header-linkage PR would be the wrong place for that decision. Happy to send it as its own PR if that reading is right.

### On re-running

I have not spent a re-run. The failure is deterministic and reproduces locally on an unmodified `main` checkout, so a re-run would only confirm what the two-worktree diff already establishes.

---

PR #484, opened 2026-09-04
SEA0: directional vessel/IMU response and response-weighted moments

mgrouch:
Continues the SEA3 directional-sea route from `main` after #481. Research question: **can a conservative directional vessel/IMU response family over the deployed finite band turn the three-partition spectrum into bounded response-weighted moments, and what does the deployed source path then actually see?**

This closes the "current blocker" the ledger named at #481 — the directional response — at the continuous-time steady-state level, and replaces it with two sharper obligations. Replay free, non-promoting: nothing here promotes SEA0, P2, P3, P4 or P5.

## Response family

An envelope, not a hull. With the lever arm disabled, angular motion produces no specific force at the IMU, so only translational response enters:

```
|G_j(omega,theta)| <= rho(omega) omega^2 u_j(theta),  u = [1, |cos theta|, |sin theta|]
rho(omega) = rho_max min{1, (omega_L/omega)^p}
```

The roll-off is finite-waterline Froude–Krylov length averaging; `rho_max = 2`, `p = 1`, `L_ref = 4 m` (a *shorter* reference hull raises `omega_L` and so assumes less).

## Analytical results

- **Modulus majorant.** For `|G_j| <= m_j`, `v* Phi_u v <= |v|^T M |v|` with `M = int m m^T E dtheta`, so `lambda_max(Phi_u) <= lambda_max(M)`. Cross-axis coupling survives as off-diagonal entries; the source is never scalarized per axis.
- **Directional factorization.** The theta integral separates into constant Gram matrices `Q_r(beta_r, s_r)`, strongly non-diagonal (normalized off-diagonals to 0.998) and direction selective — a head sea loads surge 0.81 against 0.19, a beam sea reverses it.
- **Exact leak inversion.** The deployed `omega^2 = (sigma_v/sigma_eta)^2 - lambda^2` equals `E_mu[omega^2]` **exactly, for any input spectrum** — an identity, not a narrow-band approximation. Equivalently the deployed `T_z` is the moment period of the true elevation spectrum after a fourth-order high-pass at `lambda` and the response weight `rho^2`.
- **Mixture convexity.** That mean square is convex over partitions, so the extremes of the whole three-partition class occur on one-partition seas: the fifteen-parameter period enclosure collapses to a `(T_p, gamma)` screen.
- **Finiteness ladder.** Proxy elevation/velocity and band-passed acceleration moments are finite for any `p >= 0`; the raw acceleration moment needs `p > 0`; order `n` needs `n < 2p`. The deployed path needs no order above zero.

Screens over the declared domain: `0.6564 <= Tz_src/Tp <= 0.8861` (wider on both sides than the surface screen), `0.924 <= Tz_src/Tz_eta <= 1.115`, induced tuning channel `0.0644 .. 0.5051 Hz` inside the committed `0.03 .. 1.2 Hz`.

## Two obligations, reported rather than tuned away

**Sampling supplies no band limit.** With a flat response the folded specific force above Nyquist is logarithmically divergent, and not by a negligible amount: three unmodelled decades already fold in `83 m^2/s^4` (`9.14 m/s^2`), more than the whole sigma clamp interval. Under the declared roll-off the same fold is `2.36e-4`. A certified vessel/IMU roll-off is therefore a **mandatory** SEA0 assumption; the 5 ms sampling rate may not be offered in its place.

**The sigma clamp rail is reachable.** Shipping code commits `sigma_aw = min(0.9 sigma_a, 4 m/s^2)`, so the coordinate is contained *by saturation* and no sea can violate it. But an admissible crossing sea — 1.83 m at `Tp = 4.59 s` on 8.18 m at `Tp = 10.9 s`, `Hs = 8.38 m`, each partition below its own breaking steepness — gives `sigma_a = 4.66 m/s^2` against the `4/0.9 = 4.44` saturation point. On the rail `sigma_aw` is constant, so the sea-to-source map is not injective and SEA3 cannot prune the P2 language there.

The sharper fact: neither frequency-source mode reaches the rail at rest — `3.91 m/s^2` at each sea's own deployed period, `4.06` at the fixed 0.2 Hz prior. It is reached only at intermediate references, i.e. on the **estimator-lag branch between** the two source modes #481 established. The two SEA0 subcertificates meet exactly there.

Failure analysis is recorded in the ledger per the AGENTS.md protocol: failure class is a source-language *structure* failure, not a theorem, enclosure or implementation failure; it invalidates the assumption that a physical sea domain prunes the tuner language *uniformly*; it leaves the theorem, the response family, the majorant, the factorization, the leak identity and the period channel intact.

## Domain declarations

`Tp in [2.5, 20] s`, `gamma in [1, 7]`, `Hs <= 8.5 m` (matching `significant_wave_height_Hs_upper_m` in the declared operating domain — deliberately not narrower), spreading `s in [1, 25]`, `beta` free, significant steepness `<= 0.10` imposed **per partition and on the combined sea**. The per-partition form matters: without it the search returns a 0.28-steepness chop riding on a long swell, which is not a sea. A test asserts the domain is not narrower than the declared deployment envelope.

## Honesty of the numerics

Fixed-grid Simpson quadrature with analytical tail pads, not validated interval integration, and the channels are continuous-time steady state rather than the exact discrete moment EMAs, log-period state and `AdaptiveWaveBandPass` recursion. Every numerical interval is marked a screening enclosure. Verified against closed forms where they exist: `m0 = H^2/16` and the PM `Tz/Tp = ((5/4)pi)^(-1/4)` to 9+ places, the `s = 1` Gram entries `0.5` and `2/pi` exactly.

## Changes

- `tools/ou3_sea3_directional_response_moments.py` + committed JSON artifact (pure stdlib; the `sea0` CI job installs nothing)
- `tests/validation/test_ou3_sea3_directional_response_moments.py` — 15 tests
- `ou3-proof.yml` `sea0` job: compile, `--check`, run, assert, upload
- `doc/kalman_ou_iii/w3d-sea3-directional-response.tex-part`, included after the period bridge; cleanup contract updated
- `docs/ou3-proof-research-state.md`: new subcertificate, failure analysis, rewritten blocker and next increment, four new forbidden shortcuts
- `ou3_sea3_spectral_moment_bridge.py`: one public `jonswap_dimensionless_shape` wrapper so both SEA0 artifacts share one JONSWAP convention (its own artifact is unchanged)

`ruff check tools/ tests/validation/` clean; SEA0 and contract tests pass locally. No `.c/.cc/.cpp/.h/.mk/Makefile` change, so no `make all` rebuild is implied.

## Next research question (separate PR)

Decide the rail — certify a translational-RAO bound putting `rho_max` below 1.907, or carry the saturated cell explicitly unpruned — then replace the continuous-time weights by the exact discrete estimator/band-pass recursions, declare the sea rate relation `R_lambda`, and attempt `Lhat_SEA3 subset L_current_source`.

https://claude.ai/code/session_013MnEc72HhZDeCjfQuEnrGh

---
_Generated by [Claude Code](https://claude.ai/code/session_013MnEc72HhZDeCjfQuEnrGh)_

### DIFF
--- a/.github/workflows/ou3-proof.yml
+++ b/.github/workflows/ou3-proof.yml
@@ -113,23 +113,30 @@ jobs:
           python3 -m py_compile \
             tools/ou3_sea3_spectral_moment_bridge.py \
             tools/ou3_sea3_wave_period_frontend.py \
+            tools/ou3_sea3_directional_response_moments.py \
             tests/validation/test_ou3_sea3_spectral_moment_bridge.py \
-            tests/validation/test_ou3_sea3_wave_period_frontend.py
+            tests/validation/test_ou3_sea3_wave_period_frontend.py \
+            tests/validation/test_ou3_sea3_directional_response_moments.py
 
       - name: Certify retained SEA0 subcertificates
         run: |
           python3 tools/ou3_sea3_spectral_moment_bridge.py --check
+          python3 tools/ou3_sea3_directional_response_moments.py --check
           python3 -m unittest -v \
             tests.validation.test_ou3_sea3_spectral_moment_bridge \
-            tests.validation.test_ou3_sea3_wave_period_frontend
+            tests.validation.test_ou3_sea3_wave_period_frontend \
+            tests.validation.test_ou3_sea3_directional_response_moments
           python3 tools/ou3_sea3_spectral_moment_bridge.py \
             --output /tmp/ou3_sea3_spectral_moment_bridge.json
           python3 tools/ou3_sea3_wave_period_frontend.py \
             --output /tmp/ou3_sea3_wave_period_frontend.json
+          python3 tools/ou3_sea3_directional_response_moments.py \
+            --output /tmp/ou3_sea3_directional_response_moments.json
           python3 - <<'PY'
           import json
           spectral=json.load(open('/tmp/ou3_sea3_spectral_moment_bridge.json'))
           frontend=json.load(open('/tmp/ou3_sea3_wave_period_frontend.json'))
+          response=json.load(open('/tmp/ou3_sea3_directional_response_moments.json'))
           assert spectral['validation_pass']
           assert spectral['SEA0_full_certificate_promoted'] is False
           assert spectral['trajectory_replay_used'] is False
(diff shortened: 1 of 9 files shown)


mgrouch:
Superseded by #482. Both PRs started from the same post-#481 main state, but #484 uses a different provisional response family (`rho_max=2`, `p=1`, `L_ref=4 m`) than the robust continuum now carried by #482 (`G<=4`, `p>=2`, arbitrary phase/heading/cross-axis coupling). Merging the branch wholesale would leave contradictory RAO theorem domains. The useful analytical findings from #484—response-weighted moments, the live sigma-clamp rail, alias/roll-off necessity, and the need for coupled sea/RAO admissibility—are being absorbed into #482's single canonical proof path. Closing this PR to keep one active OU-III SEA3 proof branch.

PR #488, opened 2026-09-04
Make the consumed RAO box a snapshot, and sync the SEA3 docs

mgrouch:
## What is red on main

`build / ou-evidence / commit` fails at `make -C tests/validation evidence-test`
on a single test:

```
FAIL: test_sea3_validator_rejects_mismatched_consumed_rao_box
AssertionError: 'P2 inclusion consumed a different RAO parameter box'
  not found in ['certified RAO gain range changed']
```

Same failure on `b6ea0c3` (357 tests) and on `3e14815` (361 tests). Because the
job dies before its commit step, the regenerated evidence is never published,
which is why the committed `reports/results` tree is still stamped `7890c08`
while `SeaStateFusionFilter_OU_II.h`, `SeaStateFusionFilter_OU_III.h` and
`WavePeriodEstimator.h` have moved on. That resolves itself once this job can
finish: the replay fingerprint already reports the tree stale, so `ou-evidence`
takes the regenerate-and-commit path.

The `validation: replay dependency differs from replay provenance: ...` line in
the same log is expected output from the archive-provenance tests, not a second
failure.

## Cause

`_inclusion_from_p2` stored the response-enclosure box object itself:

```python
"RAO_parameter_box_consumed": response["rao_envelope_parameter_box"],
```

so `response_enclosure.rao_envelope_parameter_box` and
`p2_inclusion.RAO_parameter_box_consumed` were one dict. `validate` compares
the two, and `copy.deepcopy` preserves aliasing within a single call, so the
comparison could only ever see an object against itself — the
`"P2 inclusion consumed a different RAO parameter box"` failure was
unreachable, and substituting a box mutated the certified enclosure instead,
tripping `"certified RAO gain range changed"`.

The producer's own docstring already claimed the artifact "records the exact
RAO parameter box it consumed, and validation requires byte-for-byte structural
equality"; the alias is what made that untrue.

## Fix

Record an independent snapshot, so the equality check is a real comparison.

## Docs the code changes never reached

`e5f035c`, `76656f5` and `139730b` landed contract strengthenings without the
matching `docs:` commit that every earlier increment on that branch had:

- the leak identity's preconditions are a finite second moment **and** a
  finite, strictly positive weighted denominator — not "finite weighted
  moments" — and source parity now pins estimator update order, not just
  fragment presence;
- the Cartesian-product refutation witness integrates `x = f/f_p` over
  `[1,9]` (docs said `[1,3]`), giving a mean-square lower bound of about
  `21.49 m^2/s^4` (RMS about `4.64 m/s^2`), and is certified to lie inside the
  declared RAO parameter ranges;
- the consumed-box binding is now stated in both SEA3 documents, and the
  aliased record is listed as a forbidden shortcut.

## Validation

- `tools/quality_gates.sh python` — ruff and `py_compile` clean.
- `tests/validation/test_ou3_source_domain_contract` — running locally; this
  module is minutes of interval arithmetic, so CI (`ou3-proof`, which owns it)
  is the authoritative verdict.

https://claude.ai/code/session_011QoQKaBmW5NjNwtV3qcobv

---
_Generated by [Claude Code](https://claude.ai/code/session_011QoQKaBmW5NjNwtV3qcobv)_

## Summary by CodeRabbit

- **Documentation**
  - Clarified proof conditions for weighted spectral identities and their distinction from continuous-time steady-state behavior.
  - Updated the incompatibility example with a broader frequency range and a certified lower-bound estimate.
  - Documented the exact response-parameter range used for SEA3-to-P2 inclusion evidence.

- **Validation**
  - Strengthened checks to reject incomplete, stale, substituted, or aliased parameter and proof records.
  - Ensured inclusion evidence is validated against an independent snapshot of the consumed response parameters.

### DIFF
--- a/docs/ou3-proof-research-state.md
+++ b/docs/ou3-proof-research-state.md
@@ -151,15 +151,30 @@ Let
 
 `W(omega)=omega^4/(omega^2+lambda^2)^4`.
 
-For **any** nonnegative input specific-force spectrum with finite weighted
-moments,
+For any nonnegative input specific-force spectrum whose weighted moments
+satisfy
+
+`0 < int W S_a < infinity`,
+
+`int omega^2 W S_a < infinity`,
+
+the identity is
 
 `sigma_v^2/sigma_eta^2 - lambda^2 = int omega^2 W S_a / int W S_a`.
 
+A finite, strictly positive weighted denominator is a precondition of the
+identity rather than an aside: the producer records it as
+`weighted_denominator_finite_and_strictly_positive_required` and the validator
+rejects an artifact that drops it. Source parity is pinned to the exact
+shipping recurrence *and* to the order in which the two high-pass stages, their
+previous-sample states, and the two leaky integrations are updated, so a
+reordered estimator cannot inherit the transfer relation.
+
 Thus the leak subtraction is exactly a weighted mean-square-frequency identity;
 it is not a narrow-band or single-sinusoid approximation. This does not replace
-the still-open exact discrete finite-EWMA/log-period transient enclosure and is
-not itself a P2-pruning certificate.
+the still-open exact discrete finite-EWMA/log-period transient enclosure, is not
+itself a P2-pruning certificate, and does not identify the discrete estimator
+with its continuous-time steady state.
 
 ### Robust directional RAO envelope family
 
@@ -218,10 +233,13 @@ PM is used only because its normalization is closed form; this is **not** a
 PM-only theorem and no claim is made that `gamma=1` is the worst JONSWAP
 member.
 
-Using only positive spectral mass on `x=f/f_p in [1,3]`, the certificate obtains
-an outward-rounded lower bound on acceleration mean square that is strictly
-larger than `4^2`. Therefore the complete independent outer product cannot be
-promoted into P1.
+Using only positive spectral mass on `x=f/f_p in [1,9]`, the certificate obtains
+an outward-rounded lower bound on acceleration mean square of about
+`21.49 m^2/s^4`, i.e. an RMS lower bound of about `4.64 m/s^2`, strictly larger
+than the `4^2` cap. The witness parameters are additionally certified to be
+finite and to lie inside the declared RAO parameter ranges, so the refutation
+rests on an admitted family member rather than an out-of-family extreme.
+Therefore the complete independent outer product cannot be promoted into P1.
 
 The correct physical theorem domain is consequently a **coupled SEA3 set**:
 sea and RAO parameters are not independently maximized, and a physical pair is
@@ -247,6 +265,11 @@ P2.
 The result is deliberately **non-pruning** and still contains all 800 current P2
 physical tuner cells. That is sound right inclusion, not yet a source
 reachability reduction.
+
+The inclusion artifact records the exact RAO parameter box it consumed as an
+independent snapshot of the response-enclosure box, and the validator requires
+structural equality between the two. A stale or substituted box is therefore
+rejected rather than inherited.
 
 ## Current certificate status
 
@@ -324,6 +347,8 @@ Do not reintroduce:
 - an independent Cartesian product of sea extremes and RAO-envelope extremes
   as if it automatically satisfied P1;
 - independent Cartesian `tau/sigma/R_S` extrema;
+- an aliased rather than snapshotted record of the RAO parameter box consumed
+  by the P2 inclusion, which makes the box-equality check vacuous;
 - replay-selected sea/source words as a certificate domain;
 - equating physical `T_p` with deployed `T_z` or averaging modal periods;
 - unbanded JONSWAP surface-acceleration variance;
--- a/docs/ou3-sea3-directional-p2-ha.md
+++ b/docs/ou3-sea3-directional-p2-ha.md
@@ -131,6 +131,12 @@ has yet removed a P2 cell.  A robust RAO range is now part of the theorem, but
 finite-memory WavePeriodEstimator/variance/tuner reachability is still required
 before the response family can be used for source-history pruning.
 
+The inclusion artifact nevertheless records the exact RAO parameter box it
+consumed as an independent snapshot of the response-enclosure box, that is of
+the declared continuum of section 1, and validation requires structural
+equality between the two.  A stale or substituted box is therefore rejected
+rather than inherited.
+
 ## 5. H/A feasibility
 
 The H/A check remains the unchanged source-complete P2-V1 translation route,
@@ -179,4 +185,5 @@ source-domain test already run there.  The retired dedicated
 CI rejects regression to a nominal RAO, a finite sampled RAO catalogue,
 independent Cartesian response boxes, response roll-off weaker than `p=2`,
 independent maximum `H_r/T_p,r` corners, three independently maximized partition
-heights, or a changed `1e-18` H/A gate.
+heights, a P2 inclusion that consumed a different RAO parameter box than the
+certified response enclosure, or a changed `1e-18` H/A gate.
--- a/tests/kalman_ou_iii/kalman_ou_iii-sim.cpp
+++ b/tests/kalman_ou_iii/kalman_ou_iii-sim.cpp
@@ -1003,12 +1003,33 @@ private:
 // over eight
 // records and three seed triplets the retune moves yaw +0.06% and pitch
 // +0.42%, against 3D -3.33% with every one of the 24 cells agreeing in sign.
+//
+// Re-cut once more for the measured-period Live entry.  Holding the fixed
+// wave-band prior until WavePeriodEstimator's startup-usable gate clears,
+// instead of taking that estimator's frequency as soon as it is finite, delays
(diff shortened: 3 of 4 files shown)


mgrouch:
`build (kalman_ou_iii)` is failing on this branch, and it is not this PR's failure.

```
Processing ./wave_data_pmstokes_H4.000_L112.766_A30.00_P30.00.csv (type=pmstokes)
ERROR: Pitch RMS above limit (0.201886 deg > 0.2007 deg). Failing.
```

This PR changes three files — `tools/ou3_sea3_directional_p2_ha_feasibility.py`
and two `docs/*.md`. `git diff --name-only origin/main...HEAD` lists exactly
those. No C++, no simulator input, no Makefile: the tree this gate compiles and
runs is byte-identical to `main`, so the same run on `main` produces the same
number.

### Why main looks green on this gate

It does not run there. In `build.yml` the `build` job requires
`needs['ou-evidence'].result == 'success'` on `refs/heads/main`, and
`ou-evidence` has been failing since #482, so every `build (…)` job on main has
been skipped rather than executed.

The change that moved this number is #486 (`f346b80`, "startup: enter live on
measured wave period"), which made `tuner_frequency_hz_()` hold the fixed
wave-band prior until `wave_period_.hasUsablePeriod()` clears instead of taking
the estimator's value as soon as it is finite, and added the same gate to the
`TunerWarm -> TunerReady` transition. That delays the tuner reaching the true
operating point, which is consistent with a small attitude-accuracy cost on the
worst case.

#486's own `build` run ([33906424361](https://github.com/bareboat-necessities/ocean-imu/actions/runs/33906424361))
executed **zero jobs** — it was superseded 21 seconds after it started, and the
merge landed at 18:32:39. The last `build` that actually completed on that
branch ([33906285720](https://github.com/bareboat-necessities/ocean-imu/actions/runs/33906285720),
success) was at `a859073`, the commit *before* the filter change. So this gate
has never run against `f346b80` until now.

### Status

The headroom here was always thin — `tests/kalman_ou_iii/kalman_ou_iii-sim.cpp`
carries `err_limit_pitch_deg = 0.2007f, // was 0.1975, worst 0.1996 (pmstokes
H4.0)`, so 0.201886 is about 1.1% above the previously recorded worst case.

I am not porting a fix into this PR. Raising the limit to pass is gate tuning,
which `AGENTS.md` forbids, and recovering the accuracy means revisiting #486's
startup gating — a behaviour decision for its author, not something to fold into
an unrelated proof-tool fix. No fix exists to port yet.

Everything else on this head is green or still running: `python`, `cppcheck`,
`sanitizers (ASan + UBSan)`, `sea0`, `p2-correlation-interface`, and every other
`build`/`build-MCU` target have passed. `source-foundation`, which runs
`test_ou3_source_domain_contract` — the test this PR actually fixes — is still in
progress; locally that module is `Ran 11 tests in 912.459s / OK`.

---

mgrouch:
Following up on my previous comment: at the repo owner's direction the pitch
gate is re-cut in this PR after all, in `189a0a5`.

I also need to correct that comment. I wrote that raising the limit "is gate
tuning, which `AGENTS.md` forbids". That was too strong. The forbidden-shortcuts
list sits inside the OU-III **proof** research protocol and targets proof
thresholds — the `1e-18` H/A gate, operating-domain shrink, deployed-filter
retuning. These simulation bars are a different thing, and the block above
`FAIL_LIMITS` documents several deliberate re-cuts, including one that moved
pitch alone (`0.1975 -> 0.2007`) for the `R_S_xy_factor 1 -> 0.72` retune.

### What was measured

The simulator already has `W3D_COLLECT_ALL_GATES`, which reports every violated
bar instead of exiting on the first, so one failure cannot mask another. On the
protocol the file documents — default seeds, the eight reference records,
trailing 900 s — the worsts are:

| bar | worst | limit | |
|---|---|---|---|
| `err_limit_pitch_deg` | 0.201886 | 0.2007 | **exceeds** |
| `err_limit_roll_deg` | 0.346879 | 0.3493 | holds |
| `err_limit_yaw_deg` | 0.899665 | 0.9023 | holds |
| `err_limit_percent_z_jonswap` | 4.4684 | 4.489 | holds |
| `err_limit_percent_z_pmstokes` | 4.44143 | 4.462 | holds |
| `err_limit_percent_3d_jonswap` | 12.7081 | 12.77 | holds |
| `err_limit_percent_3d_pmstokes` | 13.3522 | 13.43 | holds |
| `acc_z_bias_percent` | 4.28073 | 4.3 | holds |
| `bias_3d_percent` | 78.7447 | 79.0 | holds |
| `gyro_bias_3d_percent` | 15.6784 | 15.73 | holds |

Nine of ten hold, so only pitch moves — a single-bar re-cut, not a re-gauge. The
local run reproduces CI's `0.201886` exactly, so the number is not runner- or
`-march=native`-specific.

The new value follows the block's own stated half-percent/four-significant-digit
rule: `0.201886 * 1.005` rounds up to `0.2029`. I checked that rule against six
of the previously recorded re-cuts before applying it. Rebuilt and re-ran
fail-fast as CI does: all eight records `QUALITY_GATE: PASS=1`, exit 0.

Nothing outside the simulator mirrors this constant, so the change is confined
to `tests/kalman_ou_iii/kalman_ou_iii-sim.cpp`.

### Still worth knowing

The accuracy cost itself is real and is now recorded rather than hidden: #486's
startup gating costs +1.1% worst-case pitch on pmstokes H4.0. If that is not an
acceptable trade, the alternative is revisiting `tuner_frequency_hz_()` rather
than the bar — that is a separate decision, and this PR does not foreclose it.

Separately, `build-MCU (full_marine_ins)` failed once on the previous head at
`arduino-cli core install esp32:esp32@3.3.7` with "Download failed: context
canceled" — a toolchain download, before anything compiled — while the same job
passed on the push run of that identical commit. This push re-runs it.

---

mgrouch:
A third check is red on `189a0a5`: `validate` in the `ou-validation` smoke run
([33919356008](https://github.com/bareboat-necessities/ocean-imu/actions/runs/33919356008)).
It is the committed-evidence contract, and it fails on `main`'s tree
independently of this PR. **I cannot make it green from inside this PR — merging
this PR is what fixes it.**

```
make -C tests/validation test
python3 ../../tools/ou_evidence_contract.py --auto
validation:  replay dependency differs from replay provenance: src/kalman_ou_ii/SeaStateFusionFilter_OU_II.h
validation:  replay dependency differs from replay provenance: src/kalman_ou_iii/SeaStateFusionFilter_OU_III.h
validation:  replay dependency differs from replay provenance: src/tuner/WavePeriodEstimator.h
validation:  replay dependency differs from replay provenance: tests/kalman_ou_iii/kalman_ou_iii-sim.cpp
robustness:  ... OU_III.h, WavePeriodEstimator.h, kalman_ou_iii-sim.cpp
make: *** [Makefile:22: evidence-contract] Error 1
```

### It fails on clean main

Checked out `origin/main` in a separate worktree, with none of this PR's commits
present, and ran the same contract:

```
validation: replay dependency differs from replay provenance: src/kalman_ou_ii/SeaStateFusionFilter_OU_II.h
validation: replay dependency differs from replay provenance: src/kalman_ou_iii/SeaStateFusionFilter_OU_III.h
validation: replay dependency differs from replay provenance: src/tuner/WavePeriodEstimator.h
robustness: replay dependency differs from replay provenance: src/kalman_ou_iii/SeaStateFusionFilter_OU_III.h
robustness: replay dependency differs from replay provenance: src/tuner/WavePeriodEstimator.h
```

The committed bundles record `git_commit` `7890c081…`. #486 then changed those
three headers without regenerating evidence, so the recorded replay dependency
hashes have not matched the source tree since it merged.

To be straight about my own contribution: `189a0a5` adds
`kalman_ou_iii-sim.cpp` as a fourth entry, and changing a `.cpp` under `tests/`
is what made the smoke job run at all — on the previous head it was correctly
skipped. But the failing condition is the three pre-existing headers; removing
my commit changes the entry count, not the verdict.

### Why merging fixes it

The only thing that can refresh `replay_provenance` is a full simulator
regeneration, and `ou_evidence_provenance.py` deliberately refuses to stamp it
from anything else — restatement must first prove the replay-producing
implementation still matches byte-for-byte. That regeneration is the `commit`
job in `ou-validation.yml`, which is main-only and re-records all four files at
once.

That job has been failing since #482 on
`test_sea3_validator_rejects_mismatched_consumed_rao_box` — the test `47e7bea`
in this PR fixes. So the loop is:

- `ou-evidence` on main is red → evidence is never regenerated → committed
  provenance stays stale → `validate` is red on any PR that touches a
  replay-closure source file;
- `ou-evidence` is red **because of the aliasing bug this PR fixes**.

Merging this PR lets `ou-evidence` reach the regeneration step on main, which
commits fresh bundles and clears all four entries. I cannot shortcut it here:
regeneration is roughly 840 validation replays plus ~310 robustness ones,
sharded across hours, and it commits to the branch it runs on.

The evidence fingerprint gate already agrees this is stale — `replay_required`
is true for this head, so main will regenerate rather than try to reuse the
committed rows.

---

PR #501, opened 2026-09-07
OU-III P4: connect same-history nonlinear graph lineage

mgrouch:
## Phase 2 research question

Can the branch-correlated complete-SEA3 prefix selectors from #500 be consumed directly by the retained exact nonlinear prediction/Joseph/reset/projection machinery for both H18 and A21, while preserving the same event-local P/H/R cells, actual-applied anisotropic R_S, and one same-history A21 absolute accelerometer-bias lineage?

This is a stacked, non-promoting continuation of #500. Base is `codex/ou3-p4-same-history-prefix-selectors`.

## What this phase adds

- `ou3_p4_same_history_nonlinear_graph_lineage.py` consumes one explicit root-to-endpoint selector lineage.
- Prediction uses only that selector's provider body-rate coordinate, shipping sample interval, and committed tau.
- Every Joseph event consumes its passive P-before/H/R cell captured inside the same trusted shipping transition.
- Every due S=0 retains `ACTUAL_APPLIED_SPECTRALMSE_RS` provenance.
- A21 composes the deployed 0.4 m/s^2 bias projection after every Joseph/reset and fails closed unless one absolute physical-bias history covers every and only the same selector lineage.
- The smoke uses the identically-zero physical bias path as one homogeneous finite-tau_b **point history only**. It is not a source-uniform bias box and cannot promote P4.

## Hard boundaries

- Canonical source remains `COMPLETE_SEA3_NORMAL_LIVE_WORD`.
- Frozen conditional P3 delta remains `1e-18`.
- No filter, quality-gate, or declared-domain change.
- No replay-derived theorem source, independent event boxes, independent true-bias boxes, selected-S word, or state elimination.
- P4 and P5 remain open.

## Falsification / next gate

This phase must first pass the full existing P4 discovery suite. The next theorem step is to materialize the correlated finite-window complete-SEA3 source cells, including the same-history absolute A21 physical-bias coordinate, then lift this exact nonlinear lineage into dense joint graph sectors. Only after that source connection exists will we attempt the canonical-point full augmented LDLT. If that point master is not strictly feasible, the current storage route stops before source-uniform covering.

### DIFF
--- a/docs/ou3-proof-research-state.md
+++ b/docs/ou3-proof-research-state.md
@@ -179,3 +179,15 @@ The master and graph-sector machinery are non-promoting. P4 remains open until
 the source-uniform endpoint and every-prefix augmented LDLT plus domain
 retention close on the same complete SEA3 history. P5 remains blocked until
 strict canonical P4 contraction closes.
+
+## PR #500 experiment 1 failure analysis — arbitrary two-prefix covariance fixture
+
+Hypothesis: the trusted typed complete-SEA3 execution kernel can be instrumented passively to retain branch-correlated every-prefix ancestry and event-local H18/A21 Riccati cells (P-before/P-after plus the exact F/Q, floor increment, or Joseph H/R) without reimplementing shipping transitions or losing actual-applied R_S provenance.
+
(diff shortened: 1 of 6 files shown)


mgrouch (commented):
Consolidated into PR #500. All commits from this branch are now contained in #500; closing this stacked PR to keep a single canonical continuation.

PR #503, opened 2026-09-07
OU-III SEA3: add correlated finite-window source cells

mgrouch:
## Phase 3A research question

Can we implement the refinement/cover object needed for source-uniform P4 without Cartesianizing the complete SEA3 source or independently splitting derived shipping coordinates?

This is a stacked continuation of #501/#500 and remains non-promoting.

## What this adds

- `Sea3WindowCell`: one constrained 3 s / 601-transition cell inside `COMPLETE_SEA3_NORMAL_LIVE_WORD`.
- Binary refinement is intersection of the same parent source cell with one generator-coordinate halfspace; every common hard-realization, lambda-coupling and joint-response constraint is retained on both children.
- Refinement is allowed only on genuine source/generator coordinates (`lambda`, `x^s`/hard-realization, response, eventual physical-bias coordinates).
- Derived shipping coordinates such as tau, sigma_aw, R_S, T_S, P/H/K/F/Q, tuner/front-end outputs cannot define independent refinement axes.
- One window source cell can be attached to every prefix of one #500 selector lineage.
- Optional A21 absolute physical-bias intervals must cover every and only the same selector prefixes.

## What remains deliberately false

- hard SEA3 realization/membership oracle
- joint translational/rotational source-output oracle
- hard source-uniform physical A21 bias path
- source-uniform complete-window cover
- nonlinear graph-sector closure
- endpoint/prefix LDLT and P4

The existing source layer already shows why these stay false: continuum phase propagation is closed, but the hard spectral driver, physical left inclusion, and joint source-output map are not yet executable. This PR does not replace those obligations with per-sample norm hulls or a finite frequency grid.

## Next gate

Implement/validate the common-witness hard-window output oracle needed to evaluate a `Sea3WindowCell`, then connect its emitted coordinates to the nonlinear graph lineage. P4 remains open and P3 remains frozen at 1e-18.

### DIFF
--- a/tests/validation/test_ou3_sea3_correlated_window_cells.py
+++ b/tests/validation/test_ou3_sea3_correlated_window_cells.py
@@ -1,0 +1,98 @@
+#!/usr/bin/env python3
+import sys
+import unittest
+from pathlib import Path
+
+ROOT = Path(__file__).resolve().parents[2]
+sys.path.insert(0, str(ROOT / "tools"))
+sys.path.insert(0, str(ROOT / "tools" / "stability"))
+
+from ou3_interval import Interval
+import ou3_sea3_correlated_window_cells as CELLS
+
+
+class Sea3CorrelatedWindowCellsTests(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.payload = CELLS.build()
+
+    def test_status_is_structural_and_nonpromoting(self):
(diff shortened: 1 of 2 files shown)


mgrouch (commented):
Consolidated into PR #500. The Phase 3A commits are now on #500's branch; closing this stacked PR so #500 is the single continuation.

PR #524, opened 2026-09-12
OU-III ALT: compose Live magnetic calibration, interleaving and timing proof

mgrouch:
## Research question
Continue the independent OU-III ALT finite-runtime stability proof for the CURRENT SHIPPING IMPLEMENTATION while preserving the original P2/P3/P4/P5 track unchanged and independently continuable.

## Final PR #524 checkpoint

Canonical handoff for the next conversation: `docs/ou3-alt-pr524-handover.md`.

PR #524 materially advances the finite-runtime product. In addition to the previously retained source/startup/magnetic/tuner/prediction machinery, the branch now binds the machine coefficient history to the actual admitted frontend path:

- shipping float API dt/gyro/accel boundary;
- persistent AccelVibrationGuard;
- common private-Mahony vertical observer;
- tracker LPF and stillness driven from the same vertical output;
- separate/FMA band/statistics/sigma histories;
- staged TuneState and active tau/Sigma_aw/R_S/cadence;
- machine prediction displacement, historical pending a_w floor, covariance hygiene, pseudo-S scheduler/service;
- machine `Racc` vibration/RAO/restore path derived from the same guard+tuner history;
- propagation of machine-vs-exact coefficient displacement through the accelerometer measurement path;
- a joined wrapper requiring the frontend/stillness source and Racc/TuneState event to be the same already executed admitted event.

Runtime ancestry is tightened: the machine guard configuration is the binary32 image of the carried runtime guard config; tracker LPF cutoff is tied to the shipping default constructor (`MAX_FREQ_HZ`, currently 6 Hz); and the private-Mahony adapter uses `vertical_cfg.{two_kp,two_ki,gravity,settle_sec}`. The earlier erroneous `vertical_cfg.accel_gate` reference is fixed on the final branch.

## Validation / fail-closed state

The earlier full-suite failure that reported `vertical_cfg.accel_gate` came from a stale head; the final branch contains the `settle_sec` correction. Current workflows for the final handoff commit must be read separately.

Inherited theorem/source blockers remain intentionally fail-closed, including continuous/private-Mahony all-Live invariant/startup-seed closure and shared source/Riccati PE refinement obligations. Do not weaken those prerequisites to obtain a green suite.

No storage/rho search was used as a shortcut. No replay/seed fitting, independent coefficient boxes, covariance-consistency entry assumptions, physical-domain shrinking, wordwise S reset, or exact-shadow historical substitution was introduced.

`ALT_LIVE_PASS=false`, `ALT_STARTUP_PASS=false`, `ALT_END_TO_END_PASS=false`, `storage_search_allowed=false`.

The original proof track, shipping implementation, quality gates, and frozen `P3=1e-18` remain unchanged.

## Next continuation

After merge, sync latest `main`, read `AGENTS.md` and `docs/ou3-alt-pr524-handover.md`, then continue ALT in a new PR. The next priority is to carry the now-bound guard -> private-Mahony -> LPF/stillness -> band/sigma -> tuner -> Racc machine history through the actual startup/goLive path, close the admitted-source private-Mahony/startup invariant, finish deployment arithmetic/source-uniform supplies, and qualify every event of the exact 600-transition word. Only after the finite-master guard closes may common/compatible joint24 storage, every-prefix retention, and the ultimate bound be attempted.

### DIFF
--- a/.github/workflows/ou3-alt-contraction.yml
+++ b/.github/workflows/ou3-alt-contraction.yml
@@ -119,6 +119,9 @@ jobs:
             test_finite_tuner_stillness_projection test_finite_tuner_candidate \
             test_finite_tuner_projection_bridge test_finite_tuner_frontend_prefix \
             test_finite_tuner_frequency_binary32 test_finite_tuner_tau_binary32 \
+            test_finite_shipping_tau_target_binary32 test_finite_tuner_tau_roundoff_bound \
+            test_finite_tuner_tau_600_domain test_finite_complete_word_tau_qualification \
+            test_finite_tuner_tau_deployment_ledger \
             test_finite_tuner_commit test_finite_tuner_boundary_commit \
             test_finite_runtime_parameters test_finite_active_runtime_word \
             test_finite_live_imu_prefix test_finite_tilt_watchdog \
--- a/tests/ou3_alt_contraction/test_finite_admitted_disturbance_imu_word.py
+++ b/tests/ou3_alt_contraction/test_finite_admitted_disturbance_imu_word.py
@@ -35,16 +35,14 @@ class Tests(unittest.TestCase):
 
     def test_small_symbolic_bound_rejects_actual_executed_supply(self):
         s=root(0); witness,segment,raw,r,b,dynamic=BASE.operands(s.live)
-        # Perturb the same-packet accelerometer residual, not gyro.  This keeps
-        # the attitude prediction on the original branch while still making the
-        # executed ISS forcing nonzero, so this regression isolates the carried
-        # disturbance-bound guard rather than requiring unrelated trig witnesses.
-        residual=(F(1),0,0)
-        raw2=replace(raw,accel_residual_internal=residual,
-                     raw_accel_body=(raw.raw_accel_body[0]+1,raw.raw_accel_body[1],raw.raw_accel_body[2]))
+        # Exercise a forcing coordinate that is explicitly retained by the
+        # executed-word supply layer but cannot change attitude/prediction
+        # topology: shipping k_a*(T-35 C).  At T=36 C the thermal coordinate is
+        # nonzero, so a theorem history with W=0 must reject the actual event.
+        args=PBASE.root_args(); args['temperature_c']=F(36)
         with self.assertRaisesRegex(ValueError,'exceeds carried theorem bound'):
-            X.imu_step(s,restricted=r,bias_restricted=b,witness=witness,raw=raw2,
-                       packet_id='imu-1',**PBASE.root_args(),**dynamic)
+            X.imu_step(s,restricted=r,bias_restricted=b,witness=witness,raw=raw,
+                       packet_id='imu-1',**args,**dynamic)
         self.assertEqual(len(s.live.live_word.source.steps),0)
 
     def test_readiness_does_not_promote_storage(self):
--- a/tests/ou3_alt_contraction/test_finite_admitted_tau_interleaved_prefix.py
+++ b/tests/ou3_alt_contraction/test_finite_admitted_tau_interleaved_prefix.py
@@ -17,23 +17,21 @@ import test_finite_source_bound_prediction_word as PBASE
 
 
 def qualified_prefix():
-    p=BASE.begin()
-    source=p.prefix.live.live_word
+    p=BASE.begin(); source=p.prefix.live.live_word
     source=replace(source,runtime=QBASE.shipping_runtime())
     admitted=replace(p.prefix.live,live_word=source)
-    interleaved=replace(p.prefix,live=admitted)
-    return replace(p,prefix=interleaved)
+    return replace(p,prefix=replace(p.prefix,live=admitted))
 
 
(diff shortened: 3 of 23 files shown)


mgrouch:
## ALT continuation — `2eb143dcc8d490a10a64f9e38a90523521fb3510`

### Mathematical advance
The source-bound finite word now carries an executable coupled acceleration-moment prefix graph, not just source/generator labels. For adjacent durations `a,b`, the proof retains `J_ab=T(b)J_a+J_b` and proves `G(a+b)=T(b)G(a)T(b)^T+G(b)`. The exact nonnegative projection-loss identity yields

`E(J_prefix,H) <= sum_i E(J_i,h_i) <= A_max^2 H`,

with all nine moment coordinates sharing the same three-axis acceleration budget. The prefix moments and budget are derived from the physical segments actually consumed before each IMU event; no independent supply, S restart, or gain replay is introduced. The full derivation is in `docs/ou3-alt-source-continuation.md`.

The same constructor now checks existing physical p/v/a/S vector caps, the coupled moment IQC, and necessary projective rotation / physical sampled-rate constraints. All BIAS0/1/2 variants are exercised separately; one actual phi persists at fixed 5 ms, including BIAS2's phi=1 endpoint. Driver and true-beta component/norm bounds are retained, and existing outward floating endpoints are converted exactly.

This also fixes a real source-qualification overclaim: `Qualified*` names and matching history strings did not establish COMPLETE-BRMM/O^601 membership. Readiness and handoffs now explicitly distinguish checked necessary outer constraints from the still-open generating-history admission. Regression inputs that exceed a physical cap or have an impossible independent-sign moment combination are rejected; they are not described as admitted shipping instability counterexamples.

### Repairs and validation
- Fixed source-word fixtures using nonexistent `delta_theta/delta_velocity/delta_position`; they now consume the actual `J0/J1/J2` fields.
- All **511 tests** in the updated focused finite-identity workflow selection pass locally, with `OU3_ALT_REQUIRE_NATIVE=1`.
- Native shipping finite-identity correspondence passes: observed/plain sample states are bit-identical and tracked shipping headers unchanged.
- Native preserve-yaw reset regression passes: quaternion max error `5.96046448e-08`, intermediate-axis covariance max error `2.38418579e-07`, cross-covariance and near-parallel identity errors zero. These are regression tolerances, not universal disturbance bounds.
- The independently testable finite-storage rejection barrier passes without asserting legacy source closure. Full legacy `phase1_closure.build/validate` still execute every prerequisite; the full inherited suite is **not** claimed passing. Its source-building path exposed the unchanged `float(mom['A_max_mps2'])` failure on exact string `'44/5'` in `ou3_p4_brmm_physical_acceleration_witness_sector.py`.
- Required `make all` was attempted. It fails at `g++ ... -I/usr/include/eigen3 ... -c ahrs-qmekf-sim.cpp`, with `src/ahrs/KalmanQMEKF.h:30:10: fatal error: Eigen/Dense: No such file or directory`. Native checks used an existing separate Eigen installation; no full-build pass is inferred.
- The pushed Git tree `27e9d16a87804bdaacc9e826532fde30a9cb776f` exactly matches the locally validated tree. Original non-ALT research-ledger text, shipping implementation, original P2/P3/P4/P5 code and P3=1e-18 remain unchanged.

### Remaining boundary
Full generator/potential/QO inclusion and BIAS driver/parameter histories, remaining same-source runtime arithmetic/solver/roundoff branches, sample-zero magnetic binding, startup/ungauged capture, continuation beyond the 600-transition container, and clock/counter lifetime remain open. Necessary moment constraints are not a full source-uniform shipping word. Common joint24 storage, every-prefix chart retention and an ultimate bound are still unproved.

No storage/rho search was run. `ALT_LIVE_PASS=false`, `ALT_STARTUP_PASS=false`, `ALT_END_TO_END_PASS=false`. PR remains draft; newly triggered CI must be checked separately from these local results.

mgrouch:
ALT continuation update (head b7d741f):

- Verified the branch already contains the stronger whole-machine TuneState startup/goLive/Live product (tau + sigma + R_S, common pending boundary, dual compiler histories, applied ActiveParameters, and 600-step carrier).
- Found and corrected a real same-sample ordering bug in `finite_machine_active_prediction_roots`: when a pending tuner boundary fires, prediction uses the **post-boundary** exact ActiveParameters, but the old relation always formed the exact-vs-machine join against the stale pre-boundary `WORD.State.active`.
- `finite_machine_active_prediction_roots.build(...)` now accepts an explicit executed `exact_active` and forms the supply against that object; the historical no-boundary/component behavior remains available when omitted.
- Added a regression showing that an executed post-boundary exact active state is retained in the join instead of silently reverting to the predecessor active state.
- No storage/rho gate was changed. `machine_root_effect_injected_into_joint24_event_relation`, scheduler/R_S machine effects, source-uniform coefficient bounds, and all ALT PASS gates remain false.

Next proof step: make this corrected post-boundary machine prediction-root attachment mandatory on every admitted Live IMU edge (both separate/FMA histories), then attach pseudo-period scheduler and R_S S-service effects before attempting any source-uniform coefficient bound or storage search.

PR #532, opened 2026-09-17
Fix the red quality-gates and evidence-test CI failures

mgrouch:
Four CI jobs were red on `main` for reasons unrelated to the OU-III proof track. Each is fixed at its cause. No proof gate, threshold, or qualification flag is touched.

## quality-gates / warnings (`-Werror=maybe-uninitialized`)

`reference` and `frozen` in `startup_antialigned_docking.cpp` are written at `k==dock` before any read, but GCC cannot see that across the loop at `-O2`. Zero-initialized; when `count<=dock` they are never read either way.

## quality-gates / clang-tidy

- **`cert-err34-c`** — `std::atoi` cannot report a conversion error, so the sample count now goes through `std::strtol`.
- **`bugprone-suspicious-memory-comparison`** — `float` has no unique object representation, so `memcmp` over the observer state is a defect in general, even though bit-exactness is what this diagnostic means. The comparison now reads the bit patterns out explicitly, keeping the fixed-run semantics exactly as they were.

## quality-gates / python (ruff, 11 findings)

Unused imports and dead local bindings. Two calls are kept *without* their bindings because they are checks rather than values: `S.run` uses `check=True`, so disassembling the linked ELF verifies it. The `finite_wpe_log_binary32` import in `finite_master_guard` went with the dead bindings that were its only users — the guard still reads that module's state through `finite_wpe_machine_binary32`.

## build / ou-evidence / commit (evidence-test)

`test_ou3_brmm` expected `physical_constants["S_m"]` to be `None`. The BRMM declaration has bound `S_m` to the centered-primitive upper bound since *"proof: bind BRMM contract to padded physical condition"*, which postdates the test, and `S_primitive_contract` states that `S_m` is a qualified consequence of that primitive rather than an independent S ball. The test now checks `S_m` against the primitive the way it already checks `V_m`, `P_m`, and `A_m`, instead of listing it as an open obligation.

## Verification

Run locally the way `tools/quality_gates.sh` runs in CI:

- warnings gate clean over all 65 translation units
- clang-tidy clean
- ruff clean
- the 60 ALT tests covering every edited module pass
- `tests/validation` evidence list: the failing test now passes and no new failure appears

## Not in scope

The `ou3-alt-contraction` workflow is also red, with 5 failures and 10 errors that are open proof obligations rather than CI hygiene (unclosed fresh-entry cover, gain-outer covariance envelope, `phase1_storage_search_allowed_consumed`, an `AD derivative dimensions differ` defect, and a brittle `0.49999999999999983 != 0.5` endpoint assertion). Under AGENTS.md those belong to the research state machine and to a PR that declares them as its research question — not to this one. Branch `claude/ou3-alt-stability-proof-x2rz83` already addresses several of them, including the AD width defect.

https://claude.ai/code/session_015ufrYekxRC82ZmqujFdxMQ

---
_Generated by [Claude Code](https://claude.ai/code/session_015ufrYekxRC82ZmqujFdxMQ)_

## Summary by CodeRabbit

* **Tests**
  * Improved floating-point state comparisons and command-line count parsing in stability tests.
  * Updated validation expectations for derived physical values.
  * Removed unused test imports.

* **Refactor**
  * Simplified internal qualification, tuning, and runtime modules by removing unused imports, variables, and readiness checks.
  * Preserved existing qualification, validation, and fail-closed behavior.


mgrouch:
## Nine checks red on head `5646f47` — all pre-existing on `main`, none caused by this PR

Every failure below was reproduced on the base commit `e02eec9` in a clean worktree with **no part of this PR applied**. They fall into three distinct root causes.

| check | workflow | root cause |
|---|---|---|
| `physical-source` | `ou3-brmm-physical-source` | **A** |
| `relation` | `ou3-p4-joint-estimator-relation` | **A** |
| `construction` | `ou3-source-cover-construction` | **A** |
| `same-history-execution` | `ou3-proof` (×2 runs) | **A** |
| `Finite physical identities` | `ou3-alt-contraction` (×2 runs) | **A** (red on `main` @ `e02eec9`, run 35161855897) |
| `validate` | `ou-validation` | **B** |
| `riccati-p3` | `ou3-proof` | **C** |

This PR touches nine files — `startup_antialigned_docking.cpp`, `test_ou3_brmm.py`, `test_finite_mahony_prefix_totality.py`, and six modules under `tools/stability/ou3_alt_contraction/`. **It touches no `src/` file and no module in any failing call chain.** These workflows fired only because their triggers glob `tools/stability/**`, which the `ou3_alt_contraction/` edits match — the trigger matched, the cause did not.

---

### Cause A — unclosed fresh-entry obligation

`tools/stability/ou3_p4_hard_entry_set.py:29`, reached via `ou3_p4_complete_brmm_source_cover_contract`:

```
RuntimeError: qualified fresh-entry source failed: [
  'all_hard_entry_coordinates_inside_full_scale not true',
  'source_uniform_reachable_other_entry_coordinates_closed not true',
  'fresh_live_entry_cover_closed not true',
  'one or more coordinate memberships failed']
```

On clean base `e02eec9`:

```
$ python3 tools/stability/ou3_p4_complete_brmm_source_cover_contract.py --output /tmp/scc.json
RuntimeError: qualified fresh-entry source failed: [...]      exit=1

$ (cd tests/validation && python3 -m unittest test_ou3_p4_source_indexed_phi_rebase)
FAILED (errors=1)                    # same message

$ (cd tests/validation && python3 -m unittest test_ou3_mems_bias_contract \
     test_ou3_p4_same_history_nonlinear_graph_lineage ... test_ou3_p4_complete_brmm_differential_events)
Ran 50 tests in 237.487s   FAILED (errors=1)      # same message
```

`fresh_live_entry_cover_closed` is an **open proof obligation**, not a hygiene defect. Closing it is mathematics. AGENTS.md routes that through `EXECUTE → FAILURE ANALYSIS → REPLAN` with a ledger entry, forbids "artificial operating-domain reduction to obtain PASS", and asks one research question per PR. Making it green here would mean weakening a proof assertion.

`claude/ou3-alt-stability-proof-x2rz83` (PR #533) is on the ALT track and does fix the related `AD derivative dimensions differ` width bug in `ou3_p4_complete_brmm_differential_events.py`, but touches **none** of the fresh-entry chain, so there is nothing to port. `construction` is red on that unrelated branch too (run 35169638140), which corroborates base-redness.

### Cause B — stale replay provenance

`validate` fails in `tools/ou_evidence_contract.py --auto`:

```
validation: replay dependency differs from replay provenance: src/kalman_ou_iii/Kalman3D_Wave_OU_III.h
validation: replay dependency differs from replay provenance: src/kalman_ou_iii/SeaStateFusionFilter_OU_III.h
validation: replay dependency differs from replay provenance: src/tuner/MagAutoTuner.h
robustness: ... (same three)
make: *** [Makefile:26: evidence-contract] Error 1
```

Identical on clean base `e02eec9` (`exit=1`). This PR modifies **zero** `src/` files, and none of those three headers — verified with `git diff --name-only e02eec9 HEAD -- <those paths>` (empty). The recorded provenance is stale against the current headers; regenerating it is a separate change.

### Cause C — stale constant, and the repo already contradicts itself

`riccati-p3` fails on a genuinely different cause:

```
FAIL: test_height_period_coupling_blocks_unphysical_rectangular_corners
  File "tests/validation/test_ou3_brmm_physical_admissibility.py", line 56
    self.assertEqual(c["total_Hs_upper_m"], 8.5)
AssertionError: 9.35 != 8.5
```

Identical on clean base. **9.35 is the intended value** — `tests/validation/test_ou3_source_domain_contract.py:37` asserts the opposite of line 56:

```python
self.assertEqual(d["repository_total_Hs_upper_m"], expected); self.assertGreaterEqual(expected, 9.35)
```

Two tests in this repo disagree, and the other one requires `>= 9.35`. This is the same staleness class this PR already fixes for `S_m`: a hardcoded mirror left behind by a deliberate contract change (the padded envelope, `8.5 × 1.1 = 9.35`).

**Proposed patch** — ties the mirror to its source so it cannot go stale again, exactly as this PR does for `S_m`:

```diff
--- a/tests/validation/test_ou3_brmm_physical_admissibility.py
+++ b/tests/validation/test_ou3_brmm_physical_admissibility.py
@@ -53,7 +53,7 @@
         self.assertTrue(c["independent_H_r_and_T_p_rectangular_extrema_forbidden"])
         self.assertTrue(c["independent_three_partition_H_maxima_forbidden"])
-        self.assertEqual(c["total_Hs_upper_m"], 8.5)
+        self.assertEqual(c["total_Hs_upper_m"], d["repository_total_Hs_upper_m"])
```

Verified on base: `test_ou3_brmm_physical_admissibility` → `Ran 3 tests OK`, and `test_ou3_source_domain_contract` still → `Ran 9 tests OK`. (The other `8.5` on line 50 is a `partition_admissible` input, correctly left alone.)

I have **not** pushed this. It is a base-branch failure in a file this PR's scope does not cover, and the guidance is to propose rather than widen the PR. Say the word and I will add it here, or it can go in its own change.

---

### Scope of this PR

Deliberately limited to the four CI failures that were **not** proof obligations: the three `quality-gates` jobs (warnings, clang-tidy, ruff) and `build` / `ou-evidence / commit` (`evidence-test`). Those are the checks that should go green on this head; both were still queued at the time of writing and I will report their real results.

`physical-source` has been re-run once to confirm; that re-run is spent, and the others share established root causes, so I have not re-run them.

---

PR #547, opened 2026-09-18
Correct a wrong claim in the rho anisotropy handover plan

mgrouch:
Documentation-only follow-up to #545, which merged before this correction was ready.

## The wrong claim

`docs/ou3-rho-anisotropy-proof-plan.md`, as merged, said three modules read the deployed factor out of the header and need no edit:

> `ou3_brmm_riccati_tube.py:254`, `ou3_source_reachable_matrix_p3.py:247` and `ou3_brmm_tuner_scheduler_step.py:227` regex the deployed value out of the header.

Only the third is true. The other two parse it correctly and then **fail to certify**:

```
ou3_brmm_riccati_tube.py:254
  RuntimeError: cannot certify scaled OU process cell
                [0.009999999068167651, 0.010000000000000037] at depth 20
ou3_source_reachable_matrix_p3.py:247
  RuntimeError: cannot certify scaled OU process cell
                [0.00041666665735344007, 0.00041667021815050694]
```

A continuation working from the merged plan would have treated both as free and been surprised.

## Attribution: pre-existing

Both builds were run at `238a70e`, the commit before the `rho` split, where the header still carried isotropic `rho_y = 0.72`. **Both fail there with byte-identical cell intervals.** The obstruction is in the scaled OU process cell and is independent of the horizontal factor, so it is pre-existing and separable from the `rho` work.

Neither module is reached by `test_ou3_brmm_riccati_tube`, which exercises `ou3_brmm_riccati_tube_factored` — a different module that still closes and is also what the two class B P4 modules import as `TUBE`. That is why no workflow is red because of either failure, and why CI could not answer the attribution question.

## Changes

- New class D section, "Pre-existing certificate failures, not caused by this change", carrying both errors, the reproduction commands and the `238a70e` evidence.
- Class C reduced to `ou3_brmm_tuner_scheduler_step.py:227`, with a note distinguishing `ou3_brmm_riccati_tube_factored` from the unfactored module.
- Work item 3 changed from "settle the class D attribution before anything else" to "leave class D alone; it is its own change", so it cannot block items 1 and 2.
- Acceptance criterion changed to "class D is unchanged", so the continuation is not asked to close a certificate that was never this line of work's to close.

No code, no gate, no proof module changes. `P4_PASS`, `P5_MAY_START` and every `ALT_LIVE_PASS` stay false.

https://claude.ai/code/session_01DE2GwXkafDh4B8Y8Dff1VM

---
_Generated by [Claude Code](https://claude.ai/code/session_01DE2GwXkafDh4B8Y8Dff1VM)_

## Summary by CodeRabbit

* **Documentation**
  * Clarified the classification of certificate-module impacts for the horizontal anisotropy change.
  * Documented the two pre-existing module failures, including their error messages, reproduction commands, and historical attribution.
  * Updated work items and acceptance criteria to distinguish blocking and non-blocking validation work.
  * Added details about validation-gate results and clarified when regenerated evidence becomes current.


mgrouch:
CI status on `2278cea`, and why this PR is not the cause.

**`physical-source`** — three failures, two of which are not this PR's:

```
ERROR test_status_preserves_compact_brmm_and_fails_closed
ERROR test_exact_rebase_algebra
```

Both reduce to `R_S source parity failed: horizontal_RS_factors_are_0p72` via `ou3_brmm_correlated_window_outer_enclosure.py:63`. This reproduces on base `87449cb` — `ou3_brmm_complete_source.validate()` returns exactly those two strings there. It is the class A breakage that #545 shipped deliberately and that this PR's own document records; no fix exists yet, because generalising the parity premise is the work the plan hands off.

**`construction`** — the same premise, same call path. Not this PR's, same reason.

The third `physical-source` failure was real and was mine:

```
FAIL: test_counter_reaudit_preserves_all_other_wrapper_source
AssertionError: 'f4ba0350…' != '1008e931…'
```

That one is a regression from #545 — it passes at `4886aa6` and fails at `87449cb`. `test_ou3_brmm_infinite_continuation.py` pins the header twice, directly and as a counterfactual reconstruction, and #545 re-pinned only the direct hash. Fixed in #548, which takes the `physical-source` step from 3 failures to 2.

**`build (kalman_ou_iii)`** — pre-existing, and the accel bias gates that motivated #545 are intact. Built and run against the pinned v1.2.1 dataset on a tree whose C++ is byte-identical to `main`:

```
SIM EXIT=0
8 QUALITY_GATE lines, all PASS=1
```

The job still reds because `run_tests.sh` runs under `bash -e` and `live_entry_audit-test` fails before the script reaches `exit "$sim_status"`:

```
FAIL: quiet pair executes actual H18 to A21 release
FAIL: reachable session-origin S at quiet handoff exceeds the old 300 assumption
LIVE_ENTRY_AUDIT_PASS=false
```

That is the known live-entry failure from #542, unrelated to the `rho` work and untouched by it.

I have deliberately not ported #548's one-line change into this PR. It would not make this PR green — the parity premise fails regardless — and this PR is a documentation-only correction that is easier to review without an unrelated test edit in it. Once #548 merges, this branch picks the fix up from the base.

This PR changes one markdown file and cannot affect any certificate.

---

PR #548, opened 2026-09-18
Re-pin the counterfactual mag-counter hash broken by the rho split

mgrouch:
Fixes a test that #545 broke on `main`.

## The failure

`ou3-brmm-physical-source` is red on `main` with:

```
FAIL: test_counter_reaudit_preserves_all_other_wrapper_source
AssertionError: 'f4ba0350…' != '1008e931…'
```

**This is a regression from #545, not pre-existing.** The test passes at `4886aa6` and fails at `87449cb`.

## Cause

`test_ou3_brmm_infinite_continuation.py` pins the OU-III header **twice**:

1. `C.AUDITED[path]` — the direct SHA of the deployed header.
2. A *counterfactual* SHA of the header reconstructed with the mag-counter saturation guard replaced by the old `mag_updates_applied_++` and the `<limits>` include removed.

#545 changed `R_S_y_factor_` from `0.72f` to `0.50f` and re-pinned the first kind across six modules. It did not re-pin the second — the reconstruction moves with any header edit just as the direct hash does, and my sweep only looked for direct source pins. So #545's claim that all 84 of those tests recompute and close was wrong.

## Fix

Recompute the reconstruction against the deployed header. The new value `f4ba0350…` is exactly what both CI and a local run compute, so this is a re-pin, not a loosened assertion. Both other assertions in the test are untouched and still bind, including the direct `C.AUDITED` hash.

## Verified

`ou3-brmm-physical-source`'s test step goes from 3 failures to 2, confirmed in CI on this PR's head:

```
Ran 60 tests in 8.586s
FAILED (errors=2)
```

Before the change the same step read `FAILED (failures=1, errors=2)`.

Both remaining errors are the `horizontal_RS_factors_are_0p72` parity premise, which reproduces on base `87449cb` (`ou3_brmm_complete_source.validate()` returns those two failures there) and is recorded as class A in `docs/ou3-rho-anisotropy-proof-plan.md`.

I also ran every other test in `tests/validation` carrying a SHA pin — `test_ou3_brmm_infinite_continuation`, `test_ou3_brmm_runtime`, `test_ou_evidence_contract`, `test_svg_reproducibility` — and no other pin is stale.

## Other red checks, all pre-existing

Nine checks were investigated, each reproduced against base `87449cb`; see the comment thread for the per-check evidence. Summary:

- `diagnose`, `source-foundation`, `same-history-execution`, `metric-memory`, `construction`, `complete-brmm-source` — all the class A parity premise, identical on base and head.
- `validate` and `evidence / commit` — the stale evidence tree. **Correction to what this description said earlier:** this cannot be cleared by letting `ou-validation` regenerate. The evidence job regenerates and then runs the validation gate before committing, and that gate fails on the same parity premise (`make: *** [Makefile:32: evidence-test] Error 1`), so the commit step never runs. Work item 5 of the handover plan is blocked by work item 1; #547 records that ordering.
- `build (kalman_ou_iii)` — pre-existing, and the accel bias gates from #545 are intact. Built and run against the pinned v1.2.1 dataset on a tree whose C++ is byte-identical to `main`: `SIM EXIT=0`, 8 `QUALITY_GATE` lines, all `PASS=1`. The job reds because `run_tests.sh` runs under `bash -e` and `live_entry_audit-test` fails before the script reaches `exit "$sim_status"` — the known live-entry failure from #542.

https://claude.ai/code/session_01DE2GwXkafDh4B8Y8Dff1VM


mgrouch:
`complete-brmm-source` is failing on `0b57813`, and it is not this PR's.

The job ends with `##[error]Process completed with exit code 2` from `ou3_brmm_complete_source.py`. Running that module the way the workflow does, on this PR's head and on base `87449cb`:

```
BASE main   exit=2
PR548 head  exit=2
```

with the same reason in both:

```json
"P3_source_contract_ready": false,
"source_parity": { "horizontal_RS_factors_are_0p72": false }
```

This is the class A parity premise: `ou3_brmm_complete_source.py:103` asserts the deployed header contains `R_S_x_factor_` **and** `R_S_y_factor_` at `0.72f`, and #545 shipped `R_S_y_factor_ = 0.50f`. It was broken deliberately and is red on `main` independent of this PR.

**No fix exists yet to port.** Generalising that premise to extract both deployed factors is the handoff work described in `docs/ou3-rho-anisotropy-proof-plan.md` (work item 1), not something to fold into a one-line hash re-pin.

I have not spent a re-run on it: reproducing the same exit and the same JSON on the base branch is stronger evidence than a second run of the same job would be.

This PR changes one literal in one test. What it does fix — `test_counter_reaudit_preserves_all_other_wrapper_source` in `ou3-brmm-physical-source` — is verified separately in the description.

---

mgrouch:
Follow-up covering the remaining failing checks on `0b57813`. All of them are red on base `87449cb` as well. This PR changes one string literal in one test file.

I reproduced each locally against both the base and this PR's head rather than inferring the cascade.

**`diagnose` and `source-foundation`** run the same 15-test list. Identical result on both sides:

```
BASE     rc=1  test_ou3_source_domain_contract
HEAD548  rc=1  test_ou3_source_domain_contract
```

**`same-history-execution`** — identical, six for six:

```
BASE and HEAD548 both:
  test_ou3_p4_same_history_nonlinear_graph_lineage
  test_ou3_brmm_correlated_window_cells
  test_ou3_p4_complete_brmm_same_history_prefix_selectors
  test_ou3_brmm_complete_window_execution_kernel
  test_ou3_p4_complete_brmm_finite_map_mean_value
  test_ou3_p4_complete_brmm_differential_events
```

**`metric-memory`** — identical on both, for the test and the module it builds:

```
BASE     test rc=1, module rc=1
HEAD548  test rc=1, module rc=1

RuntimeError: four-S complete-BRMM prerequisites failed:
  {'complete': ['P3_source_contract_ready is not true',
                'R_S source parity failed: horizontal_RS_factors_are_0p72']}
```

**`construction`** — the parity premise reaching it through `ou3_brmm_correlated_window_outer_enclosure.py:63`, the same path already shown for `complete-brmm-source` above.

**`validate` and `evidence / commit`** — one knot, and also downstream of the same premise.

`validate` fails identically on both sides:

```
BASE     FAIL: test_committed_bundles_match_immutable_replay_provenance
HEAD548  FAIL: test_committed_bundles_match_immutable_replay_provenance
```

#545 moved two replay dependencies, `src/kalman_ou_iii/SeaStateFusionFilter_OU_III.h` and `tests/kalman_ou_iii/kalman_ou_iii-sim.cpp`, and the last regeneration on that branch ran at `238a70e`, before the filter change landed in `0cb5ab0`. So the committed bundles describe the previous filter.

The obvious remedy — let the evidence branch job regenerate and commit — does not currently work, and this is worth flagging because it is not obvious from the plan. That job regenerates and then runs the validation gate before committing, and the gate fails on the parity premise:

```
FAIL: test_complete_brmm_conditional_source_does_not_claim_physical_left_inclusion
  (test_ou3_source_domain_contract)
AssertionError: ['P3_source_contract_ready is not true',
                 'R_S source parity failed: horizontal_RS_factors_are_0p72'] != []
Ran 499 tests - FAILED (failures=1, skipped=1)
make: *** [Makefile:32: evidence-test] Error 1
```

So the commit step never runs and the tree stays stale. The evidence cannot be refreshed until the parity premise is generalised — work item 5 of the handover plan is blocked by work item 1, not independent of it. #547 records this ordering in the plan so a continuation does not start with item 5 and hand-edit the evidence tree to force it.

**`physical-source`** — this is the one this PR acts on. In CI it now reports:

```
Ran 60 tests in 8.586s
FAILED (errors=2)
```

Before this change the same step read `FAILED (failures=1, errors=2)`. The `failures=1` was the counterfactual-hash regression from #545; it is gone. The two remaining errors are the parity premise.

So across nine failing checks there was exactly one failure attributable to anything other than the deliberate class A premise, and this PR fixes it. Everything else resolves once work item 1 lands.

No re-runs spent: reproducing each failure on the base branch settles attribution more directly than re-running the same job would.

---

PR #551, opened 2026-09-18
Record the two relation producers unmasked by the R_S parity fix

mgrouch:
Documentation only. Follow-up to #549.

## What this records

Fixing the horizontal `R_S` parity premise in #549 let the `ou3-p4-joint-estimator-relation` producer chain run past the correlated outer-enclosure prerequisite that had been aborting it. Two producers then fail on their own terms:

```
ou3_p4_brmm_event_lineage_cover.py:167   ValueError: detached event token
ou3_p4_brmm_a21_projection_prefix.py:54  same-Dtheta A21 correction target
                                         did not close, 26411 attempts
```

Neither is caused by #549 or by the `rho` split in `0cb5ab0`. The premise failure was simply aborting the chain before either was reached, so fixing it makes them visible.

## Attribution

Settled by the method already used for class D — run the producer at the commit before the header moved:

| Commit | Header | `event_lineage_cover` | `a21_projection_prefix` |
|---|---|---|---|
| `238a70e` | `0.72 / 0.72` | `detached event token` | `closed=False`, 26411 attempts |
| `0c69413` | `0.72 / 0.50` | (blocked by the premise) | `closed=False`, 26411 attempts |
| current | `0.72 / 0.50` | `detached event token` | `closed=False`, 26411 attempts |

Byte-identical at `238a70e`. The a21 chart matches in all ten reported fields (34 → 31 dimensions, 3 reset-defect columns removed, 7 hard and 11 nonlinear premises, 7 equality-like, 2 chord, 2 cross) under both the isotropic and the anisotropic pair, so that certificate is not `rho`-sensitive at all. The token guard at `source_cover_contract.py:75` is a string comparison with no numeric input.

## Why a record rather than a fix

`AGENTS.md` makes the failure analysis mandatory, and its one-research-question-per-PR rule keeps each of these out of the parity work:

- The token mismatch is a lineage-token question: establish which token the smoke cell should carry.
- The a21 non-closure is a correction-domain question and falls under the two-strike rule, so it needs a failure analysis of the limiting sector before another delta-grid or multiplier-grid widening is attempted.

No proof code, no theorem gate and no deployed constant moves. `P4_PASS`, `P5_MAY_START` and every `ALT_LIVE_PASS` stay false.

## Verification

`tools/quality_gates.sh python` clean; `test_ou3_proof_cleanup` and `test_ou3_stability_package_integrity` pass (19 tests).

https://claude.ai/code/session_01LodZszHLP6tqkaYERaEPBx

---
_Generated by [Claude Code](https://claude.ai/code/session_01LodZszHLP6tqkaYERaEPBx)_


mgrouch:
`build (kalman_ou_iii)` shows as failed against this PR. It is not this PR's failure, and main already carries the fix — no action needed here.

**Which run it is.** The failed check is [job 105466159019](https://github.com/bareboat-necessities/ocean-imu/actions/runs/35301489648/job/105466159019), in run `35301489648` — the `refs/pull/549/merge` run at head `7875245`. It is attributed here only because #549 merged, its branch was deleted, and this PR reuses the same branch name. This PR's own `build` runs are 13574 (push) and 13575 (pull_request) at `79b1429`.

**Root cause.** LaTeX, not C++. In the first compile step:

```
! Undefined control sequence.
l.181 \part
           {Assumptions, objectives, and the selected proof architecture}
! Emergency stop.
!  ==> Fatal error occurred, no output PDF file produced!
```

`doc/kalman_ou_iii/kalman_ou-w3d-stability-study.tex` uses `\part` under `\documentclass[conference]{IEEEtran}`, which does not define it.

The visible tail is from the retry step, which reports something less obvious:

```
Latexmk: Nothing to do for 'kalman_ou-w3d-stability-study.tex'.
Collected error summary: lualatex: gave an error in previous invocation of latexmk.
```

Worth noting for the workflow itself: because the first step is `continue-on-error: true` and the retry runs in the same working directory, latexmk finds the poisoned `.fdb_latexmk`/`.log` state and refuses to redo the target. The retry can therefore never recover a genuinely broken document, and it hides the real error behind an "up-to-date" message. Adding `-gg` (or `-f`) to the retry's `args`, or cleaning the aux files in its `pre_compile`, would make the retry meaningful and surface the true error in the failing step.

**Attribution.**

| Revision | `\part` in that file | LaTeX build |
| --- | --- | --- |
| `87449cb` (merge before #546) | file does not exist | `build (kalman_ou_iii)` passed |
| `0c69413` (#546 merged; base of #549) | 5 | broken |
| `41bfce3` (#549 merged) | 5 | broken |
| `80dbd74` (#550 merged; current main) | 0 | fixed |

The file was added by `585bf42` in #546 and the `\part` usage was removed by `f687fd6` ("docs: remove stability TOC and forced part page breaks") in #550. Neither #549 nor this PR touches any file under `doc/` or any `.tex` — #549 changed `tools/stability/`, `tests/`, two workflows and `docs/*.md`; this PR changes one Markdown file.

**Why nothing is ported.** The fix already exists on the base branch: this PR is based on `80dbd74`, which carries `f687fd6`, so its own runs compile the corrected document. Re-running the stale `refs/pull/549/merge` run would still fail, because that ref's tree predates #550 — it would tell us nothing about this PR.

---

mgrouch:
Second stale check, `evidence / commit` ([job 105480336329](https://github.com/bareboat-necessities/ocean-imu/actions/runs/35301444591/job/105480336329), run `35301444591` at head `7875245`) — again the #549 branch run surfaced here by the reused branch name. Not this PR's, but it carries a result worth recording, and it will reproduce on this PR because the base has it.

**The good part.** The evidence regeneration is no longer blocked:

```
Ran 509 tests in 22.230s
FAILED (failures=2, skipped=1)
```

509 tests rather than 499, and the `R_S source parity failed: horizontal_RS_factors_are_0p72` failure is gone, so `ou_evidence_contract` passed and the regenerated tree was validated. That was the premise #549 fixed.

**What still blocks the commit step.** Failures unrelated to `R_S`, all in documentation contracts left behind by the stability-study split. Reproduced on current main (`80dbd74`):

1. `test_ou_stochastic_concentration.test_conclusion_no_longer_lists_concentration_as_future_work`

   Asserts `"Conditional Gaussian quadratic-form concentration"` and `"Bernstein/Freedman"` appear in `w3d-conclusion-summary.tex-part`. `585bf42` rewrote that conclusion, and neither string exists anywhere under `doc/` now. The `assertNotIn("Sharper stochastic concentration")` half still holds; the two `assertIn`s pin deleted text. Since the text is gone rather than moved, the test cannot be repointed at the companion study — the assertions need dropping or restating against whatever now carries that claim. That is a content call for whoever owns the split, not a mechanical fix.

2. `test_ou3_stability_document_split.test_study_uses_same_ieee_two_column_style_as_article`

   This one is an over-escaping bug in the test, not a document problem:

   ```
   AssertionError: '\\documentclass[conference]{IEEEtran}' not found in
                   '\documentclass[conference]{IEEEtran}...'
   ```

   The expected string carries two backslashes; the file has one. Dropping the doubled backslash in the expected literal fixes it.

3. `test_ou_legacy_law_cleanup.test_publication_tree_has_no_legacy_numerical_law_studies`

   ```
   AssertionError: 'OU--III tuning sensitivity' unexpectedly found in
     '% Generated by tools/ou_validation.py from the committed full-study rows...'
   ```

   The generated table still carries the legacy caption, so either `tools/ou_validation.py` or the committed full-study rows still emit the retired study.

CI reported two failures at that commit; I reproduce these three locally on `80dbd74`, so the exact pair depends on the tree.

**Not fixed here.** None is related to this PR's research question, which is recording the two unmasked relation producers, and `AGENTS.md` keeps one research question per PR. (1) and (3) are content decisions inside document work that is still in flight, and I am not going to make those calls unilaterally. (2) is a one-token test fix and could ride a small cleanup PR alongside them. Happy to open that if wanted.

Worth noting these are what now keeps the regenerated OU evidence from being committed on main, so they are not cosmetic.

---

mgrouch:
Pushed `5607ab0`, which widens this PR beyond the documentation record. Flagging that explicitly since it changes the PR's scope, and it is easy to split out if you would rather it landed separately.

**What it fixes.** #549 removed the shared prerequisite failure that was aborting two proof jobs early. They now execute the searches they were skipping and overran budgets that had only ever been measured against the aborted runs. Neither job has a failing test — both were cut off mid-step, which GitHub reports as `cancelled`.

At `41bfce3`:

| Workflow / job | Budget | Observed | Siblings |
| --- | --- | --- | --- |
| `ou3-alt-contraction` / Full inherited ALT regression suite | 45 min | step ran 03:05:39 → 03:50:33, **44m54s**, unfinished | other 2 jobs passed |
| `ou3-proof` / `riccati-p4` | 30 min | job ran 03:30:03 → 04:00:19, **30m16s**, cut off inside "Test complete-BRMM finite physical P4 contract" | other 5 jobs passed |

`ou3-proof` reports 6 jobs and 0 failed jobs, which is consistent with a budget overrun rather than a test failure. For contrast, at `0c69413` both workflows concluded `failure` in about ten minutes, because the premise aborted 51 of 55 ALT modules and the P4 contract test never reached its body.

**The numbers.** The full ALT suite is 1391 tests, `OK (skipped=12)`, in 2272 s — about 38 minutes on a 16 GB host. The hosted runner had not finished at 44m54s, so it is at least 1.2× slower and 45 minutes leaves no margin. Both budgets go to 90, inside existing practice here: other workflows already budget 180 and 350.

One honest caveat: `riccati-p4` has never been observed finishing, so 90 is a first estimate with headroom, not a measured bound. If a run overruns it again the number needs raising — not the test weakening.

No test is skipped, disabled or quarantined; no proof code, theorem gate or deployed constant moves.

---

mgrouch:
`build (kalman_ou_iii)` is failing on this PR's own head (`5607ab0`), three times now, so it is not a flake and I am not re-running it again. Recording what is established and what I cannot yet see, because the evidence is genuinely contradictory.

**What fails.** Step 19, `Preserve regression and plot failures after producing review artifacts`, whose condition is

```yaml
if: steps.regression.outcome != 'success' || steps.plots.outcome != 'success'
```

so either `Run tests` or `Generate PGF plots` failed. Both carry `continue-on-error: true`, which makes their **`conclusion` read `success` in the jobs API even when they failed** — the real result lives in `outcome`, which the API does not expose. So the job API cannot tell me which one it was, and it is easy to misread this job as "everything passed".

The three failures, all at `5607ab0`:

| Run | Event | `Run tests` | `Generate PGF plots` | Result |
| --- | --- | --- | --- | --- |
| 35307510181 | push | 62 s | 16m27s | step 19 failed |
| 35307512996 | pull_request | 62 s | 16m56s | step 19 failed |
| 35307510181 (re-run) | push | — | — | step 19 failed |

Both steps run to full length in every case, so neither is bailing out early. `Compile LaTeX document` succeeds and the retry step is **skipped**, which confirms the `\part` breakage from #546 is fixed on `80dbd74` by #550.

**The contradiction.** This branch is `80dbd74` plus three files: `docs/ou3-proof-research-state.md`, `.github/workflows/ou3-alt-contraction.yml` and `.github/workflows/ou3-proof.yml`. None is `build.yml`, and none can reach the `kalman_ou_iii` simulator, its regression gates or the PGF plots. Meanwhile `build (kalman_ou_iii)` **passed** on `main` at `80dbd74` (run 35304052619, where the only failing job of 19 was `ou-evidence / commit`), and passed at `87449cb` before that. So identical code passes on main and fails three times here.

**Why I cannot name the step from CI.** `get_job_logs` is tail-only and caps at 5000 lines. The `Compile LaTeX document` step alone fills that window, so the `Run tests` and `Generate PGF plots` output is unreachable. Step 9 only copies CSVs into `plots/`; it does not upload them, so none of the run's 22 artifacts carries the gate output either.

Two things in the workflow make this harder than it needs to be, both worth fixing independently of this PR:

1. The guard collapses two different failures into one message. Reporting which of `regression`/`plots` tripped, e.g. `echo "regression=${{ steps.regression.outcome }} plots=${{ steps.plots.outcome }}"`, would make this diagnosable from the check alone.
2. Nothing preserves the regression gate output. Step 9 retains the CSVs locally but uploads nothing, so a gate failure leaves no artifact.

**Next.** I am reproducing locally (`make ensure-sim-data`, compile, then `tests/kalman_ou_iii/run_tests.sh` with `W3D_COLLECT_ALL_GATES=1`) to get the gate output directly. I will not propose a fix for a step I have not seen fail.

---

PR #558, opened 2026-09-19
Prove time-varying A21 detectability and local nonlinear stability

mgrouch:
## OU-III stability proof — handoff

This PR contains the single current proof architecture:

`construction -> finite capture -> finite H18 bridge -> finite reference refinement/release -> recurring magnetically informed A21 -> regional practical stability`.

Shipping estimator behavior and quality gates are unchanged. Retired architectures/diagnostics are not theorem paths.

### Certified/retained results

- Fixed-coordinate long-word A21 neutral information:
  `mu_N >= 2.04e-3`.
- Translation subcertificate:
  `mu_trans >= 2.04734e-3`.
- Source-uniform UCO/UCC and finite covariance compactness.
- Literal 21-state covariance/Joseph semantics, hard covariance events, verified 3x3 innovation machinery.
- Entrywise midpoint-radius Riccati enclosure was executed and retired after its second conditioning failure; do not subdivide it again.
- Covariance normalization now uses one block/factor architecture:
  AG=(attitude,gyro bias), LIN=(v,p,S,a_w), BA=(active accel bias).
- Generic absolute-cross Gershgorin was executed and retired as a conditioning reduction: it produced `gamma=-5.122292072e8`.
- Stronger root result: shipping prediction is
  `P-=F P+ F^T+Q` with block-diagonal process `Q`. Source-uniform process-factor existence therefore gives post-prediction factor-metric coercivity `gamma_root=1` independent of carried cross covariance. Finite cross-block bounds follow from covariance compactness and remain prefix/totality obligations.
- Corrections preserve positive factor-metric coercivity via
  `gamma+=gamma-/(1+gamma- j_D)`.
- Constructive AG process factor probe:
  `ell_AG=4.999974644e-4`.
- Constructive active-BA recurrence probe:
  `ell_BA=5.618273739e-4`.
- LIN scaled controllability is positive at every tested tau; the continuous-tau rigorous numerical factor remains the immediate constructive blocker.
- Accelerometer-bias projection is carried through the literal dissipative projection sector. Do **not** reintroduce projection-inactive release.

### Immediate next work on main

The next conversation should continue from current main, not reopen the retired entrywise covariance enclosure.

1. Finish the source-uniform **numerical LIN factor** over continuous `tau in [0.02,12]` and process-noise range, preserving the useful multi-step controllability factor through intervening shipping corrections.
2. Combine compatible numeric AG/LIN/BA factors with the already-closed block-factor root coercivity to obtain constructive covariance-normalized information and
   `rho0<1`.
3. Derive explicit nonlinear `r_*`.
4. Close whole-word float32 additive supply/practical radius.
5. Close finite startup/capture, H18 bridge, reference refinement/release, entry into the explicit A21 region, and every-prefix retention.
6. Finish recurring applied MAGNETIC SERVICE, physical sensor/bias/MARINE MOTION qualification, and implementation/arithmetic totality.

The theorem remains fail-closed: no end-to-end regional practical-stability claim is made yet.

### Validation

At the current PR head all required workflows are green: build, OU validation, OU-III stability proof, evidence-contract, quality gates, and stability-study rendering.


mgrouch:
Final local validation update for `ff63b0a20ddaf698a6a45e54dcaa24f3e9013b80`:

- The tested/staged source and uploaded commit have the same complete tree SHA: `1964fd43d4eefd774b4c1a0d42d011a7055ca1a6`.
- 116 OU-III Python tests and 25 shared workflow/replay-fingerprint tests passed; theorem provenance validation and Python compilation passed.
- Re-ran both shipping native regressions after the final source upload: `OU-III shipping implementation contract PASS` and `OU3_SHIPPING_TRANSITION_PASS=true`.
- The full `make all` completed compilation of every host target, then exited 2 in `ensure-sim-data`: curl could not resolve `github.com` while fetching the pinned v1.2.1 vessel-RAO archive. This is an environment/network limitation, not a completed full-suite pass. No dataset or quality gate was bypassed. PR CI has started and is the remaining full-suite validation.
- The shipping exporter and both high-precision diagnostics completed locally, including the every-prefix complement calculation. Their results remain non-promoting.
- The four-page IEEE two-column PDF was rendered and visually checked. No overfull boxes or unresolved references were reported.

The controlling proof gap is now quantitative: the finite-difference discrepancy does not fit the available contraction margin, and a same-history finite-error remainder bound is still needed. The discrepancy itself is not a rigorous uncertainty bound. Capture, source-uniform dissipativity, release/A21, recurring service, retention and arithmetic obligations remain open.

mgrouch:
Proof direction changed as requested.

The PR #557 experiment is no longer being treated as a proof track. I kept only its mathematical negative result: the held-bias identity block rules out one strict full-state H18 incremental-contraction formulation. I removed the finite-difference exporter/evaluator, H18 complement diagnostic, their dedicated tests/Makefile targets, and the generic point-witness finite-error diagnostic.

The key architecture correction is stronger: **H18 does not need asymptotic contraction at all.** In shipping code it is a finite bridge while the magnetic reference is provisional. The wrapper releases the external accelerometer-bias hold after magnetic-reference refinement; the internal gate then clears after its accepted-update threshold and one-second guard. The asymptotic proof target is therefore recurring magnetically informed **A21**.

Added analytic theorem machinery only:
- finite H18 bridge bound, valid even for one-step gain >= 1;
- finite count/guard release bound after finite reference refinement;
- A21 nonlinear small-gain lift: rho = (sqrt(rho0)+eta)^2, so the real target is sqrt(rho0)+eta < 1;
- practical-radius consequence after supply/coercivity closure;
- a sufficient finite MagAutoTuner refinement lemma tied to the literal min-sample/min-window, 35% running norm-ratio, and horizontal-field gates.

That refinement analysis exposed the next real blocker rather than hiding it: MAGNETIC SERVICE by itself does not automatically prove the MagAutoTuner's 35% running norm-ratio gate. The deployed refinement has quality weighting and startup hard-iron fitting disabled, which simplifies the state machine, but the broad 20--75 uT global field envelope is not a same-history variation bound. We now need to derive a tighter per-history corrected-field norm envelope from the physical field + residual assumptions (or explicitly qualify one) to close finite refinement/release.

After that, the main asymptotic work is source-uniform A21 linear dissipation from the literal shipping word plus MAGNETIC SERVICE, followed by a nonlinear MEKF/reset/projection/tuner remainder bound below the linear margin. No candidate numerical rho from a sampled word is being used.

mgrouch:
Further proof progress on the productive route:

1. **Finite magnetic refinement is now reduced to capture, not a new magnetic assumption.** For the deployed unweighted MagAutoTuner, rotation preserves true field norm. Existing theorem limits |B|>=20 uT and residual<=2 uT give a worst-case running norm-ratio 2R/(Bmin-R)=4/18=0.222<0.35. With horizontal field>=15 uT, |B|<=75 uT and residual<=2 uT, the literal 5% horizontal gate is guaranteed for captured tilt-frame error below about 7 degrees. Under MAGNETIC SERVICE (T_M=1 s), the 128-sample/30-s refinement therefore completes in at most 128 s after it starts. Conservatively assuming none of the internal 250-update count was earned earlier, A21 entry follows within another 251 s. The code now carries this as a captured-domain implication, not as a numerical experiment.

2. **A21 projection can be removed from the local tail nonlinearity by retention.** The estimate projection radius is 0.4 m/s^2 and the physical bias bound is 0.22516660498 m/s^2, leaving an exact inner bias-error margin 0.17483339502 m/s^2. If release retention enters that domain, projection is inactive throughout the local A21 proof.

3. **The linear A21 target is now a complete information floor, not a fitted rho.** Prediction with Q>=0 and literal Kalman corrections are covariance-metric nonexpansive. The target certificate is a normalized complete-word information floor J_A21 >= mu I; then the linear comparison is rho0<=1/(1+mu), and the nonlinear theorem closes with sqrt(rho0)+eta<1. MAGNETIC SERVICE supplies only the heading/axial-gyro-bias part; gravity/accelerometer and integral pseudo-updates must close the other directions.

4. **Translation/active-bias has no structural rank obstruction.** For scalar state (v,p,S,a_w,b_a), exact constant-parameter OU transition, four S observations and one attitude-normalized accelerometer row, a 5x5 observability minor simplifies exactly to
   det O = -Delta^3 tau^3 (1-exp(-Delta/tau))^3,
which is nonzero for every finite Delta,tau>0. The remaining task is to extend that analytic minor to the bounded time-varying shipping scheduler/tuner and combine the three translational axes with attitude/gravity and magnetic information.

No sampled candidate rho has been promoted. Current controlling work is now (a) capture into the <=~7 degree / inner-bias retained domain and finite H18 bridge retention, and (b) the time-varying complete A21 information-floor proof plus nonlinear remainder bound.

mgrouch:
Quantitative continuation uncovered and corrected an important retention issue before it could become a false theorem.

**1. Projection-inactive release is not universally reachable.** H18 holds the accelerometer-bias estimate while physical truth may be anywhere in the all-time ball B_a=0.2251666 m/s^2. Even with a zero held estimate, release error can therefore be 0.2251666, larger than the projection-inactive radius R_b-B_a=0.1748334. I removed the 0.15 m/s^2 release requirement. A21 now carries the literal projection sector ||e+||^2+||d||^2<=||e_corr||^2 and the universal post-projection radius B_a+R_b=0.6251666. This is the correct route; trying to prove startup enters the old 0.15 ball would have been another dead end.

**2. A real numerical information enclosure is now in place for the neutral translation block.** On the 16 s word, choose literal S updates after 0, 8 and 16 s. The tau-scaled scheduler guarantees <=0.156 s event delay. In normalized coordinates (v,p,S)/(5.5,8.1,1100), with declared r_S<=100, determinant + Frobenius bounds give

    mu_trans >= 2.04734e-3.

This is analytic and source-uniform; it is not a sampled singular value.

**3. I did not promote a tempting but invalid full mu_N shortcut.** A sparse two-epoch calculation gives a positive scale around 6.18e-7 and rho≈0.999999382, but MAGNETIC SERVICE is an already-transported 2-D heading/axial-gyro-bias Gramian over each service window, not an instantaneous pure-heading observation. Treating it as such would be mathematically wrong and produces a uselessly tiny nonlinear margin anyway. The full quantitative certificate must compose the actual service Gramian directly with the many recurring accelerometer rows over the 16 s word. That aggregate Schur/Gramian calculation is now the controlling mu_N task.

**4. Nonlinear and arithmetic formulas are constructive once that tight rho is enclosed.** Added explicit r*=min(r_guard,(1-sqrt(rho0))/L2) for eta(r)<=L2 r, and the additive-supply practical-radius formula. Float32 uses the rigorous kernelwise gamma_n=n*u/(1-n*u), u=2^-24. I have deliberately not claimed a whole-word arithmetic bound until literal operation counts and magnitude envelopes are composed.

**5. CI regressions from the architecture rewrite were addressed.** The Python proof test no longer uses wildcard imports (ruff failure), and the IEEE article again contains the conditional finite-error/coercivity/every-prefix publication contract expected by validation.

Next productive step is now narrow and quantitative: derive the aggregate 16-s attitude/gyro-bias Gramian lower bound from (a) every-sample accelerometer rows with |f_z|>=g-A_max, (b) the actual per-window MAGNETIC SERVICE 2-D Gramian, and (c) exact bounded-rate attitude/bias transport. That should give a materially tighter mu_N than the sparse surrogate and is the route that can plausibly leave enough 1-sqrt(rho0) margin for an explicit nonlinear radius and arithmetic supply.

mgrouch:
Quantitative continuation: I changed the proof word rather than weakening the assumptions or promoting the sparse surrogate.

A key observation is that the weak gyro-bias coordinate can be accumulated over many **disjoint one-second MAGNETIC SERVICE windows** without ever integrating attitude over a long interval. At each service root, J_hb>=I permits extraction of one unit of pure root-heading information (subtract diag(1,0); the remainder stays PSD). The simultaneous accelerometer sample has no preceding gyro-bias transport.

Using the actual shipping nominal accel std 0.2 m/s^2, vibration Racc gain 0.75, and the commissioned detector-band residual bound 0.3 m/s^2 gives a literal effective std ceiling hypot(0.2,0.75*0.3)=0.3010399. With |f_z|>=g-Amax=1.00665, |f_h|<=8.8, and heading floor 1, each root has pure attitude information j>0.0129.

For y_k=theta_0+C_k b_g, bounded angular rate gives a one-second increment singular floor
  beta = 0.02 * 2 sin(Omega_max/2)/Omega_max ~= 0.0196905.
The path-graph inequality sum|Delta y|^2 <= 4 sum|y|^2 plus y_0=theta_0 gives
  mu_ag = j*c/(1+c), c=(N-1) beta^2/4.
Choosing N=2048 disjoint service windows gives mu_ag>2.1e-3. The independent 16-s translation certificate remains mu_trans>=2.04734e-3, so the complete **fixed-coordinate neutral Gramian** now has
  mu_N >= 2.04e-3.
The 2048-s word is deliberate: it accumulates the weak physical gyro-bias scale instead of adding an excitation assumption or relying on a cancellable long attitude integral.

I also corrected an important normalization issue before turning that number into rho. The identity rho0<=1/(1+mu) is valid only for root-covariance-whitened information. A fixed physical-coordinate Gramian can be made arbitrarily large by coordinate rescaling, so directly inserting 2.04e-3 would be invalid. The theorem now explicitly requires
  P_root >= p_min I
in the same proof coordinates, giving
  mu_cov >= p_min*mu_N,
  rho0 <= 1/(1+p_min*mu_N).
This is now the controlling quantitative blocker. It is the right blocker: the classical UCO/UCC Kalman result gives existence of bounded/coercive Riccati covariance, but the remaining job is a constructive recurring lower Riccati enclosure for the literal shipping schedule.

The theorem tests, provenance validation, shipping contract and transition regressions are green on the current head; build and evidence-contract are also green. Other workflows are still running.

Next I will use the already-closed UCO/UCC constants to construct a numerical Riccati lower/upper enclosure rather than deriving p_min from a one-sample Q floor (which would be formally positive but uselessly tiny). That enclosure is what can produce a meaningful rho0 and leave enough margin for L2/r*, float32 supply, and capture/release retention.

mgrouch:
One more important advance/correction before rho0:

The weak transverse gyro-bias direction can be made quantitatively useful **without adding an excitation assumption** by lengthening the proof word. I changed the A21 theorem word to 2048 disjoint one-second service windows, while retaining the existing 16-s translation subcertificate.

At each one-second root, J_hb>=I allows extraction of one unit of pure heading information; the simultaneous root accelerometer row has zero preceding gyro-bias transport. The shipping effective accel-noise ceiling is now derived literally from the configured values:
  sigma_acc,eff <= hypot(0.2, 0.75*0.3) = 0.3010399 m/s^2.
This gives pure root-attitude information j>0.0129.

For y_k=theta_0+C_k b_g, the corrected one-second bounded-rate transport bound is
  sigma_min(C_(k+1)-C_k) >= 0.02 * 2 sin(Omega_max/2)/Omega_max ~= 0.0196905.
Using sum|Delta y|^2 <= 4 sum|y|^2 and y_0=theta_0 gives
  mu_ag = j*c/(1+c), c=(N-1)beta^2/4.
At N=2048, mu_ag>2.1e-3. The independent translation floor is 2.04734e-3, so the complete **fixed-coordinate neutral information floor is now constructive**:
  mu_N >= 2.04e-3.
This avoids the bad sparse two-epoch shortcut and never integrates attitude over 34 minutes; only one-second increments are used, so bounded rotations cannot cancel the bias information.

I also caught a coordinate-invariance issue before promoting rho0. The formula rho0<=1/(1+mu) requires root-covariance-whitened information. The fixed-coordinate mu_N above cannot be inserted directly. The correct next relation is
  P_root >= p_min I  =>  mu_cov >= p_min mu_N
  => rho0 <= 1/(1+p_min mu_N).

I added the corrected classical UCO/UCC Riccati P_min/P_max formulas as a rigorous fallback/cross-check, but I am **not** using their raw-IMU-step constants as the primary route because they are known to become excessively conservative. The primary next certificate is a verified interval Riccati enclosure over the bounded shipping schedule/hard events, with outward-rounded residual/Krawczyk verification of the innovation solves. A midpoint covariance trajectory may seed the box but cannot certify it.

This is now the actual quantitative bottleneck. Once p_min is verified, rho0 is immediate; then the already-added formulas turn a certified second-order slope L2 into r*=min(r_guard,(1-sqrt(rho0))/L2), and float32 stays as separately accumulated additive supply.

mgrouch:
Continued the p_min/Riccati blocker with executable fail-closed certificate machinery rather than promoting a midpoint covariance trajectory.

Added strict innovation-box verification: for S=S0+E, require lambda_min(S0)>=a and ||E||<=r<a; then the midpoint-inverse residual is <=r/a<1 and the certified inverse norm is <=1/(a-r). If r>=a the certificate fails closed.

Added recurring Riccati-box inclusion: an outward-rounded image R(P,parameters) must lie inside the proposed [p_min I,p_max I] box. A midpoint trajectory alone cannot satisfy this API. Certificate scope flags must all be true: full bounded shipping schedule, covariance hard events (a_w sync + bias release floor), and float32 rounding.

Also added host-independent midpoint/radius arithmetic primitives with analytic product radii plus nextafter outward expansion; float32 shipping roundoff remains a separate gamma_n budget. A committed interval-riccati-status.json is deliberately OPEN, and build_evidence.py now fails if it is ever marked complete without passing the verifier. CI tests cover strict residual/Krawczyk inverse inclusion, outward box inclusion, failure cases, JSON loading, outward dot products, and interval Gershgorin spectral bounds.

I added a conservative one-step process-noise floor primitive as a cross-check, but did not use it as the full p_min because the integrated-OU translation Gramian makes a one-step full-state bound needlessly tiny. The primary route remains the recurring interval Riccati box.

Current theorem state is accurate: fixed-coordinate mu_N>=2.04e-3 is closed; p_min is still OPEN until the full matrix image is enclosed. Once the status artifact contains a verified recurring lower bound, the certificate code automatically computes mu_cov=p_min*mu_N, rho0=1/(1+mu_cov), and the available nonlinear norm margin.

Next implementation step is the large one: encode the literal 21-state covariance prediction/Joseph maps as midpoint-radius interval kernels over dt/tau/Racc/R_S/magnetic-service parameter boxes, including the two covariance reconfiguration events, then iterate the interval map to a self-containing recurring box and verify every innovation solve. I am not substituting scalar spectral recurrences for that matrix enclosure because they lose the cross-covariance structure that makes the useful p_min possible.

mgrouch:
Implemented the next verified-Riccati layer in PR #558.

The interval proof now has literal 21-state midpoint-radius covariance kernels rather than only scalar certificate wrappers:

- P-=F P F'+Q over the full 21x21 covariance, retaining every cross-covariance;
- 3-D innovation S=H P H'+R;
- literal Joseph covariance update (I-KH)P(I-KH)'+K R K';
- interval spectral enclosure and strict innovation inverse verification;
- periodic a_w covariance-floor hard event;
- H18-to-A21 accelerometer-bias covariance-floor hard event.

I also added shipping measurement interval constructors for all three proof updates: integral S selector + R_S, magnetometer J_att + Rmag, and active-A21 accelerometer J_att/J_aw/J_ba + Racc. Lever-arm J_bg is deliberately absent because the current theorem scope has lever arm disabled. These are conservative entrywise boxes and do not themselves promote an innovation solve; the inverse still has to pass the strict residual certificate.

New CI tests verify that prediction preserves cross-covariances, the Joseph kernel matches an exact diagonal case, a wide innovation box fails closed, hard covariance events are enclosed, and the literal acc/S/mag H/R constructors cover their shipping blocks.

The committed interval-riccati status remains OPEN. It now records the exact remaining implementation gap: shipping-specific F/Q interval construction, verified interval gain solve, and iteration to a self-containing recurring 21-state covariance box. I have not substituted a scalar spectral recurrence for those pieces.

One important practical note: GitHub's structured PR metadata briefly reported mergeable=false while checks had not populated, but the direct GitHub PR endpoint reports mergeable=true / clean at the current head. New Actions runs had not appeared yet at the last poll.

mgrouch:
Continued all twelve blockers in PR #558, prioritizing the dependencies that unlock the rest.

**Root covariance / Riccati:** the generic interval layer is now complete enough to assemble the actual covariance word: source-uniform 21-state F/Q, literal S/acc/mag H/R, verified innovation inverse + interval gain, full Joseph update with every cross-covariance, AW/bias covariance hard events, and self-containing box iteration. The literal wrapper order is encoded as prediction -> optional S -> accelerometer -> optional magnetometer. I also closed the ordering lemma needed for p_min: each SPD-noise Kalman correction is Loewner-decreasing, prediction is PSD-order monotone, and the covariance hard events are PSD inflations. Therefore inserting optional S/mag corrections at every sample is a conservative **maximal-correction** cadence: its covariance is no larger than any admissible sparser shipping cadence, so a positive lower floor for it is valid for all admissible cadences.

The remaining p_min problem is now numerical/interval dependency, not missing covariance algebra: obtain a self-containing 21-state box for that maximal-correction word. If the broad entrywise attitude/parameter boxes make innovation inversion fail, the next correct move is branch subdivision/equivariant reduction, not weakening the physical theorem.

**Float32 / prefix retention:** added local-kernel gamma_n arithmetic composition propagated through certified remaining-prefix gains; this avoids an invalid single gamma_N over the entire 2048-s word. Added the corresponding every-prefix retained-set test. Numerical whole-word supply still waits for p_max/prefix gains from the verified box.

**Capture/H18/release:** added the fail-closed set-inclusion composition from a certified capture storage set through a finite H18 recurrence and literal release map into the eventual explicit A21 radius. The composition is closed; the actual capture set, history-dependent bridge length, release gains/supply and explicit r_* are still open.

**Magnetic service / physical qualification:** added an all-time MAGNETIC SERVICE continuation certificate requiring every-window same-history service, actually applied events, and actual transported innovation covariance. Finite replay explicitly cannot certify it. Added a single-history physical qualification composition requiring MARINE MOTION + IMU BIAS + MAGNETIC SERVICE to carry the same history_id plus explicit assembled sensor/mount/calibration qualification. These schemas/compositions are closed and tested; the actual all-time deployment certificates remain external/open.

**CI:** I found that the stability workflow's unittest command had collapsed onto one line earlier, so the new interval tests were not actually being executed despite prior green runs. That is fixed: CI now explicitly runs `test_ou3_interval_riccati`, `test_ou3_interval_riccati_21`, and `test_ou3_physical_qualification`. The previous proof run then correctly exposed stale provenance; I refreshed provenance after the latest theorem-source changes. New checks are starting on the current head.

The 12-blocker ledger is now committed at `docs/ou3-controlling-proof-obligations.md`, distinguishing theorem obligations from closed infrastructure. No remaining obligation was promoted merely because its helper/schema exists.

mgrouch:
Continued the remaining p_min blocker on PR #558. The current branch already had more of the literal Riccati machinery than the earlier status implied: shipping F/Q intervals, verified 3x3 interval gain solves, recurring-box iteration, literal event order, and a proved maximal-optional-correction monotonicity reduction are present. I therefore did not duplicate them.

The actual current failure mode is interval dependency: the source-uniform H/attitude/tuner boxes can become too wide for the innovation inverse and self-containing covariance box even when every point is valid. I added finite branch subdivision as the next rigorous layer. Selected dominant uncertain H/parameter entries are bisected into an exact outward cover; every branch cell must independently verify its innovation inverse. The recurring covariance image is then the hull of ALL branch images. A favorable midpoint or favorable branch can never promote the certificate.

Added `adaptive_verified_update` (subdivide only when the unsplit innovation fails), `split_interval_matrix`, and `recurring_box_over_cells`. Tests verify that split cells cover the original interval, exact cells do not subdivide unnecessarily, and one bad branch prevents recurring-box certification.

The interval-Riccati status is still OPEN and now states the precise remaining numerical task: choose a finite outward-covering subdivision of the compact shipping attitude/tuner parameter domain such that the all-branch recurring hull is self-containing and has a positive spectral floor. That is the last step before p_min can be written into the certificate and the existing mu_cov/rho0 chain can run.

Direct GitHub PR metadata reports the current head mergeable=true but mergeable_state=unstable because the new Actions runs have not populated yet.

mgrouch:
Continued PR #558 under the requested fail-closed research discipline.

I read AGENTS.md, shipping OU-III, research state, handoff, theorem status, and the stability article before continuing. I also checked the reported red stability CI: on the prior head there was no mathematical/test failure to fix; the stability-proof, validation, render, evidence and quality runs were cancelled by subsequent commits, while the full build completed successfully. New runs have not populated yet on the newest head.

For blocker 1, the branch already contained literal 21-state F/Q, Joseph, verified gain, hard events, event ordering, maximal-correction monotonicity and recurring-box iteration. The actual failed quantity was interval conditioning: the first attitude F interval enclosed R_step entrywise in [-1,1]. At Omega_max*dt_max ~=0.003666 rad this is valid but destroys essentially all near-identity structure before the first Riccati multiplication. I classified this as interval-conditioning, not covariance instability, and recorded it in the research ledger/status.

Per AGENTS.md I made exactly one mathematically justified refinement of this mechanism. The predictor now uses Rodrigues bounds |R_ii-1|<=1-cos(theta_max), |R_ij|<=sin(theta_max)+1-cos(theta_max), and a near -dt I gyro-bias injection with O(Omega dt^2) off-diagonal radius. This reduces the attitude diagonal interval radius from 1 to <7e-6 and off-diagonal radius to <0.0037. I also added finite outward-covering branch subdivision: every cell must verify its innovation inverse and the recurring image is the hull of all cells; no favorable midpoint/branch can promote p_min.

A CI/tooling bug introduced during that work was caught by inspection before CI populated: the new subdivision tests referenced helpers that were not imported. That import is fixed.

The certificate remains OPEN. The exact next falsifiable experiment is now encoded/documented: run the recurring box with the Rodrigues-tight F interval and one finite subdivision of the dominant acc/mag attitude-Jacobian entries, and record either the first failed innovation residual r/a or the recurring spectral inclusion margins. If this one refinement does not cross the threshold, the protocol requires stopping subdivision and reviewing covariance normalization rather than blindly splitting further.

Because blocker 1 has not yet genuinely closed, I have not marked blockers 2-12 complete or manufactured rho0/r*/capture claims. The retained facts are mu_N>=2.04e-3 in fixed coordinates, source-uniform UCO/UCC, literal 21-state covariance/Joseph semantics, and verified inverse machinery.

mgrouch:
Executed the required refined enclosure rather than stopping at infrastructure. The result is decisive for the current entrywise interval mechanism: on the literal 21-state prediction from the constructor covariance scales, the refined interval covariance spectral enclosure is [-18.7907040215, 2502.41441854]. The S innovation still verifies with r/a=0.66666386445<1, but accelerometer and magnetometer innovation boxes have no strict inverse certificate. This is interval dependency/conditioning, not a pointwise covariance counterexample.

Under AGENTS.md this is the second failure of the same entrywise midpoint-radius mechanism (the one permitted Rodrigues refinement was already applied), so I stopped further subdivision. The research ledger, machine status, handoff and IEEE article now record the exact failed quantities, classification, retained facts and architecture-review requirement. The old projection-inactive release wording in the handoff was also removed; the theorem carries the literal dissipative projection sector.

I then started the required PSD-structure-preserving covariance representation rather than another box tightening. The first feasibility layer uses Weyl/singular-value prediction bounds for F P F'+Q and the information-form posterior lower bound for a 3-D correction. These bounds cannot manufacture a negative covariance eigenvalue from interval dependency. A CI probe now runs prediction -> S -> accelerometer -> magnetometer on this representation and prints its positive spectral margins; it is explicitly non-promoting until a recurring source-uniform factor/spectral enclosure is obtained.

CI/tooling was also separated from mathematics. The earlier evidence-contract red was a floating-point test expectation (0.9999999999999999 vs 1.0) and then a missing test import; both are fixed. Full build and OU validation are green on the recent heads. Current newest checks are running.

Because the entrywise Riccati mechanism genuinely failed, p_min is still OPEN and I have not marked mu_cov/rho0, r*, arithmetic supply, capture, H18/release, prefix retention or totality complete. The next decision comes from the PSD-preserving probe: if its one-step margins are positive, extend that representation to a recurring factor/spectral enclosure; if it is already nonpositive, the covariance-normalization formulation itself must change before any downstream blocker can honestly advance.

mgrouch:
Continued from the latest PR #558 head after rereading AGENTS.md, shipping OU-III, research state, handoff, theorem status and the IEEE study.

Blocker 1 remains controlling, but the covariance architecture is now better characterized quantitatively. The old entrywise interval mechanism is formally retired after its permitted refinement: predicted covariance enclosure [-18.7907040,2502.41442], S innovation r/a=0.666663864 verified, acc/mag inverse unverified. I did not subdivide it again.

The PSD-preserving replacement passed its first one-step feasibility test: sigma_min(F)>=0.644087104; lower covariance floors 4.14848197e-7 after prediction, 4.14825252e-7 after S, 3.01357721e-7 after acc, and 1.49402788e-7 after mag. This proves the representation avoids the artificial negative eigenvalue, but it is not recurring p_min.

I then executed the next required process-controllability experiment instead of stopping at the representation. Two important numbers came out: with q_floor=0, the conservative PSD scalar recurrence drives a 1e-6 root floor to 1.46713043e-19 in only 32 maximal-correction samples; and the raw unscaled Gershgorin lower bound on a 0.156-s integrated-OU controllability Gramian is zero for tau=.02,.2,2,12. This is a scaling/conditioning failure, not loss of controllability: the exact integrated-OU Gramian is SPD and uniform controllability is already proved structurally.

Per the one-refinement rule, I did not keep tightening scalar norms. The mathematically motivated refinement is now implemented and running: proof-coordinate-scaled factor/LDLT on the multi-step OU controllability Gramian using the declared (v,p,S,a) scales (5.5,8.1,1100,4). It must produce positive pivots uniformly over the tau envelope before it can feed a recurring covariance certificate. If it fails, the PSD scalar/factor normalization architecture itself needs review rather than another numerical tweak.

CI status was also cleaned up. The previous quality-gates red was purely ruff: unused imports in new proof tests/tooling and one pre-existing physical-qualification test import. Those have been removed. The stability-proof/evidence/build/OU-validation workflows themselves had been green before the newest proof probe; current latest-head workflows are queued/running.

No downstream theorem obligation has been promoted. mu_N>=2.04e-3 remains the fixed-coordinate certificate; p_min, mu_cov/rho0, r*, float32 practical radius, capture/H18/release retention, prefix retention, magnetic-service completion and totality remain fail-closed until the recurring covariance floor genuinely closes.

mgrouch:
Moved PR #558 to the block/factor covariance metric, as requested. The failed entrywise midpoint-radius Riccati box is now explicitly retired after its second conditioning failure; it will not be subdivided again.

The new proof-coordinate partition is AG=(attitude,gyro bias), LIN=(v,p,S,a_w), BA=(active accelerometer bias). Each diagonal covariance block is represented by a certified factor floor P_ii >= L_i L_i^T with sigma_min(L_i)>=ell_i, while every cross covariance P_ij is retained through an operator-norm upper bound c_ij. In factor-scaled coordinates the coercivity test is gamma=1-max_i sum_(j!=i)c_ij/(ell_i ell_j)>0. The covariance-normalized information certificate is then mu_cov>=gamma*min_i(mu_i ell_i^2), followed by rho0<=1/(1+mu_cov). This avoids throwing away the useful conditioning of the OU translation block while still accounting for all cross-covariances.

Added executable block/factor theorem code and CI tests. The tests fail closed when cross coupling destroys gamma and verify that a positive block margin yields rho0<1. A non-promoting scaled integrated-OU factor probe is also in CI; it uses LDLT pivots on the scaled 4-state controllability Gramian rather than its catastrophically small raw scalar eigenvalue.

Updated theorem status, interval-Riccati status, provenance, research state, handoff and the IEEE article to make block/factor the sole covariance-normalization representation. No estimator behavior or quality gate changed.

The immediate next certificate is source-uniform rather than architectural: enclose the LIN factor over tau/process-noise variation including the within-slice controllability remainder; derive AG and BA factor floors from the literal process/release dynamics; then propagate recurring cross-block operator bounds through the actual correction/event cadence. Only a positive gamma will promote mu_cov/rho0.

mgrouch:
Handover prepared. The PR description now contains the current certified results, retired/dead-end mechanisms that must not be reopened, the block/factor covariance architecture, exact immediate blocker, and ordered continuation work for the next conversation. Current required workflows are green; proceeding to merge into main.

PR #561, opened 2026-09-20
OU-III: historical AG factor action toward the uniform six-column loss bound

mgrouch:
The controlling mathematical dependency remains open: derive quantitative bounds from the actual construction and carried nominal mean/covariance evolution that exclude sustained nominal force/field collinearity and gyro aliasing, then certify L(W)O(W)=T_h(W) and one common B_*. This PR does not claim source-uniform full-state contraction.

The single construction → capture → H18 → refinement/release → A21 proof path, estimator behavior, physical assumptions, and existing quality gates are preserved. The historical readout retains all nuisance/process correlations, accepted rank-three observations, and literal resets. The existing nuisance/Schur and six-column-to-full-contraction implications remain conditional on their unproved uniform premises.

What is established:
- The inherited BA covariance projection guard from the preceding continuation remains valid: pre-projection sqrt(V)<=6 gives zero projection defect. Entry and prefix retention are not inferred.
- Complete recorded sync/addition/symmetry operations now have exact signed-factor upper enclosures in all 21 coordinates. The moving increment has an exact -2^-44 Rayleigh defect; it is not silently treated as PSD process noise.
- Both quiet and moving exported words pass exact AG-root cancellation, full 6x6 action enclosure, and direct comparison with their literal terminal AG covariance. Prediction, solve, Joseph, reset and state-update rounding remain separate.
- Exact nullspaces record why independent nominal boxes and innovation-only reasoning cannot establish the uniform readout. The complete-turn gyro family is not proved reachable from construction or magnetically serviced and is not a shipping counterexample.

Full-construction attempt:
- Added a reproducible literal wrapper run from begin() to 600 s, with no reseeding: p=-(3/2)sin(2t)(1,0,1), R=I, B=(45,0,45), zero true biases.
- Continuous squared motion/primitive envelopes are derived analytically and satisfy the unchanged physical limits.
- The wrapper reaches Live at step 30002 and refinement/release at 36008. During 400–600 s, finite float tilt averages 8.211611 degrees (maximum 8.260681), so six-degree entry does not follow from stage flags. Nominal acceleration reaches 9.776391 m/s², invalidating substitution of the physical 8.8 bound into the nominal domain.
- Exact full-covariance SPD verification and correlated BA marginal elimination give V >= 41601169495079937572567/3619798568741048312 (>11492.6752) at the recorded endpoint. The candidate V<=36 has margin below -11456.6752. This excludes that sufficient guard at the finite endpoint; it does not prove projection is active. The evidence validator reproduces the certificate and native-driver fingerprint.
- This is not a proof of all-time magnetic service, an exact reachable-set enclosure, or a refutation of history-dependent eventual capture. Its positive finite force/field margin is not promoted to B_*.

Construction-linked mean-action continuation at a49eaa9c:
- Executed all 89998 predictions and 122251 applied rank-three corrections from construction. The six-coordinate BG/AW action retains actual cross terms and explicitly carries recorded mean roundoff. A fresh untapped control matches all terminal states, quaternion, covariance, stage times and magnetic counts.
- An 80-digit diagnostic followed by outward 40-digit intervals and independent rational interval LDL proves gyro-bias norm <1 rad/s at every recorded prefix. The corresponding increment ceiling is 0.0099051914291880918 rad.
- Exact binary-operand checks give nominal world-frame force/field sine >2/5 at all 40000 pre-accelerometer states in the recorded 400–600 s tail.
- These are finite source results. They do not establish all-time magnetic service or uniform exclusion over admitted histories.
- The proposed energy-only enclosure fails its sufficient pointwise separation inequality E_col-E>0. The exact margin lies in [-817884.048495,-817884.048494]. This is an enclosure failure, not a newly established nonphysical or reachable collinear trajectory. It does not rule out a temporal argument.
- The same-history signed innovations and physical/nominal integral relations remain the missing mathematical dependency before common B_*. No physical assumption or estimator setting was changed. CI reproduces the full transcript and its enclosures.

Uniform B_*, full covariance upper bound, J, useful rho0, nonlinear retained region, capture/entry, every-prefix retention and whole-word float32 supply remain false/open. The obstruction is mathematical, not a slow calculation. Evidence, provenance, tests, theorem status, ledger, handoff and the article state these boundaries.

Validation at a49eaa9c:
- 486 evidence/publication tests run: 485 passed, one existing data-dependent skip, against the latest automated evidence commit 62f5687b.
- Exact artifact reproduction and provenance, Ruff, Python compilation, fresh observer/control parity, the full 80-digit and outward mean-action audits, rational factor checks and the 40000-prefix exact force audit passed locally. Earlier quiet/moving exported-word enclosures remain valid.
- Nine-page draft article rendered without overfull boxes; affected pages visually checked.
- The previous required make all was interrupted at the generated-data architecture scan (exit 130); no aggregate make-all pass is claimed. This continuation changes proof tooling/docs/CI, not shipping C++.
- All six workflows passed at a49eaa9c, including full construction transcript reproduction, interval enclosure, and every quality gate. The final documentation/provenance clarification at 4917eb51 has passed local provenance validation; its CI rerun is being inspected.

The PR preserves automated evidence commits 26f6511c and 62f5687b and current main d621c016, including its draft title, release metadata and generated gallery.

Final clarification at 4917eb51: the mean action is accumulated directly from actual gains and does not rely on subtracting covariances across mean-preserving hard covariance replacements. No numerical result, executable proof code, or theorem flag changed.


mgrouch (commented):
## Signed temporal continuation at 4917eb51 — partial result, no theorem promotion

I checked the current head, AGENTS.md, controlling obligations, ledger, theorem status, historical readout/mean-action derivations, article and the shipping predictor/measurement operations. This comment is a mathematical continuation, not a code commit or a source-uniform certificate. Shipping behavior, assumptions, gates and the single proof path remain unchanged.

### Exact nominal integral identity retaining correction jumps

Use a regular window and a smooth normalized nonnegative weight psi with psi=psi'=psi''=0 at both endpoints. On each open ideal-OU prediction interpolation, Shat'=phat, phat'=vhat, vhat'=ahat. Retain literal-source/interpolation discrepancies and every actual state jump at their seams. Three integrations by parts give

```
int psi ahat dt
  = -int psi''' Shat dt
    -sum_j [psi_j Delta vhat_j
            -psi'_j Delta phat_j
            +psi''_j Delta Shat_j].
```

At an applied rank-three correction, the three jumps are the actual rows of the SAME K_j r_j. Therefore, with

```
C_psi = sum_applied_j
          (psi_j K_v,j - psi'_j K_p,j + psi''_j K_S,j) r_j,
```

the identity is `int psi ahat = -int psi''' Shat - C_psi - D_psi`. D_psi retains source-transition, arithmetic and applicable hard-event discrepancies. Frame changes require consistent transport. The identity does not turn -Shat(t) between events into additional applied pseudo-observations.

This is the necessary correction to a direct physical-to-nominal primitive transfer: measurement updates jump all three nominal integral coordinates. Physical boundedness cannot be substituted for boundedness of Shat or its signed correction charge.

For a fixed committed reference and orthonormal C with C Bhat=0, define

```
Z_psi = C [g + int psi''' Shat dt + C_psi + D_psi].
```

Then `int psi C(ahat-g) = -Z_psi`, and Jensen gives `int psi ||C(ahat-g)||^2 >= ||Z_psi||^2`. A positive lower bound on this expression over all eligible construction-carried histories would be a sufficient temporal exclusion. **No such uniform lower bound has been established here.** A vanishing weighted mean would not by itself refute a stronger temporal information bound. Full raw AG transport/reset control is still required even after this sufficient exclusion.

The signed sensor identities must remain coupled to it. With true and nominal body-to-world R,Rhat and E=R Rhat^T:

```
E(ahat-g) = a-g + R(ba+n_a-bhat_a-r_a),
E Bhat    = B + R(d_m-r_m).
```

Use each at its actual event boundary, without equating distinct acc/mag states or freezing E.

### Executed physical weighting calculation, not nominal information

The weight `psi_T(t)=140 t^3(T-t)^3/T^7` has L1 derivative norms `35/(8T)`, `336 sqrt(5)/(25T^2)` and `273/T^3`. The unchanged physical p/v/primitive bounds imply

```
||int psi_T a|| <= min(35 Vmax/(8T),
                      336 sqrt(5) Pmax/(25 T^2),
                      273 P_AC/(2 T^3)).
```

The last factor 1/2 uses equal positive/negative mass of psi''' and the pairwise physical primitive-span bound; no history reset is used. Proof-only nearest-sample weights beta_i=int_cell psi_T satisfy

```
||sum beta_i a_i - int psi_T a||
  <= Jmax [hmax/4 + hmax^2 ||psi_T'||_1/32].
```

At T=64, the existing physical limits give the exact rational sampled-mean ceiling `26808087711/128000000000 = .2094381852421875 m/s^2`. Applying only the existing TRUE-world-transport physical vector argument gives a physical 3-D Gram floor >.0002609185, including both magnetic residual bounds and conservatively rounded-up physical BA. This is NOT the nominal AG array, B_*, J, or 1-rho0. Improving it does not repair the failed unsigned-energy enclosure.

### Remaining construction-linked gyro condition

With actual correction increment u=K_bg r+d_bg, retain

```
||bhat_g+u||^2-||bhat_g||^2 = 2 bhat_g^T u + ||u||^2.
```

A source-uniform first-exit barrier must start at actual construction and retain the signed term and every seam. The recorded norm<1 result is not that barrier. Only after deriving a uniform theta=h||omega_measured-bhat_g||<=theta_*<2pi can one use the ideal gyro-integral singular bound `h sin(theta_*/2)/(theta_*/2)`, subtract its source-helper defect, and continue the complete AG analysis.

### Validation and exact stopping point

A standalone Fraction-arithmetic implementation passed 13 tests of the kernel, signed identity including all jumps/boundaries, quadrature and exact physical constants. Independent symbolic integration verified the three derivative norms. These polynomial algebra tests are NOT source executions or repository quality-gate results. No native replay, repository evidence suite or make-all pass is claimed: container git access failed with 'Could not resolve host: github.com'; connected GitHub reads succeeded. That access failure is separate from the mathematical gap.

The unresolved quantity is the common signed nominal functional/gyro barrier on actual admitted histories. Uniform LO=T_h with bounded B_W, full covariance upper bound, J, useful rho0, retained nonlinear region, capture and arithmetic totality remain open. The existing 400–600 s audits are not used as universal premises. No new reachable counterexample or stability closure is claimed.

mgrouch:
Continuation at 34d46c95 removes the dead-end formulation from the active route. The proof notes now construct the signed, chronological same-history equations: literal acc/mag innovations are substituted into the physical history; the acceleration integral is applied before absolute values; and the actual BG/AW mean recursion is telescoped with realized gains, OU transport, resets and roundoff. Sustained force/field collinearity and complete-turn gyro aliasing are reduced to source-uniform signed margins Delta_col and Delta_gyr. If both infima are positive, finite pivot-chart coverage of the separated AG observation family makes the complete historical reader action continuous and yields B_*=sup_W lambda_max(B_W)<infinity. No independent nominal box or unsigned innovation-energy implication is used.

I have deliberately NOT promoted B_* yet: the next executable obligation is the outward same-history enclosure of Delta_col and Delta_gyr over the admitted window family, followed immediately by the finite pivot-chart maximum of B_W. The existing 400–600 s audit remains diagnostic only. Shipping behavior, assumptions, gates and the single proof path are unchanged.

mgrouch (commented):
## Continuation at the new 34d46c95 head: exact adjoint requirement and an all-time capture-guard obstruction

I examined both intervening commits. They add the signed route to `docs/ou3-construction-mean-action.md` and the ledger; shipping code, physical constants and the established covariance/service lemmas are unchanged. There is still no source-uniform B_* certificate. Two further mathematical results matter before treating the remaining work as numerical enclosure.

### 1. The proposed innovation elimination requires an actual adjoint construction

For the operation-ordered recurrence

```
u_(i+1) = A_i u_i + K_i r_i + d_i,
```

let W_i be the weights obtained from the physical force/field balance and Z_i arbitrary compatible matrix multipliers. The exact identity is

```
sum_i W_i r_i
  = Z_N u_N - Z_0 u_0
    + sum_i (Z_i - Z_(i+1) A_i) u_i
    + sum_i (W_i - Z_(i+1) K_i) r_i
    - sum_i Z_(i+1) d_i.
```

Thus an endpoint-only elimination requires BOTH `Z_i=Z_(i+1)A_i` and `W_i=Z_(i+1)K_i` at every operation. Solving for a separate multiplier at each correction does not establish the inter-event adjoint recurrence. If either equality is not imposed, its interior nominal-state or innovation functional must remain and be bounded on the same history. The particular W_i produced by contracting the physical sensor relations is not automatically the coefficient family of an endpoint selector.

The earlier signed nominal-integral identity supplies one concrete compatible correction functional through the v/p/S rows, but retains its Shat bulk term and source discrepancies. It does not make an arbitrary physical innovation functional vanish. I implemented and checked the complete identity with exact Fractions. One test has zero innovations on the supplied history but nonzero coefficient mismatch, so numerical cancellation cannot be promoted to a matrix identity.

Also, positive proposed separation margins do not by themselves supply the compact coefficient/transport family invoked in the new text. A bound on the actual remaining nominal means, gains/resets and action factors, and full AG rank/endpoint control with those resets, must be established or replaced by a direct uniform action estimate. Finite operation count alone does not bound unbounded coefficients. These are proof obligations, not new physical assumptions.

### 2. Universal eventual capture into V<=36 is impossible in the current regular real-arithmetic model

This is a new quiet physical realization of the existing stationary level/north source record, NOT a revival of the inadmissible 200-Hz acceleration witness.

Let `g=196133/20000`, `c=9999/10001`, `s=200/10001`, and

```
R = [[1,0,0],[0,c,-s],[0,s,c]],
B = (75,0,0),
p = v = a = q = omega = 0 for all time,
b_g = 0,
b_a = g(R^T e_z-e_z) = (0,g*s,g*(c-1)).
```

R is a proper constant body-to-world rotation, with tilt `2 atan(.01)`, about 1.1458774 degrees. The constant physical accelerometer-bias magnitude is about .1961231941 m/s^2, strictly below .225 and hence below the unchanged .22516660498395405 limit. Bias rates, translation, velocity, acceleration, jerk and primitive span are all zero. Sensor residuals may be zero. The field meets the unchanged magnitude/horizontal limits.

Exactly,

```
R^T(-g e_z)+b_a = -g e_z,
R^T B = B,
gyro = 0.
```

The complete measured record from construction is therefore the SAME `(acc=-g e_z, gyro=0, mag=75 e_x)` used by the existing all-time source induction in `docs/ou3-sampled-capture-obstruction.md`, section 2. That induction reaches Live/refinement/release without reseeding and leaves nominal attitude I and all BG/LIN/BA means zero. Every subsequent innovation is zero. The source mean induction and magnetic-service proof depend on the measured record, not its old high-frequency physical explanation.

The existing actual-innovation service floor is exactly

```
L = 4570312500000000/1228455181462133 > 3.
```

I reproduced its exact covariance invariance and all-phase prefix comparisons. For the present true down axis `(0,s,c)`, service is at least `c^2 L`, which exceeds 2.9988 and therefore exceeds the required 1 on every tail service window. The other independent axis group adds PSD information. This is the existing REAL-ARITHMETIC source/service lemma applied to a newly admissible quiet physical history, not finite-replay extrapolation.

For the FULL SPD covariance, retaining all cross terms,

```
V=e^T P^-1 e >= e_ba^T P_ba,ba^-1 e_ba.
```

The established source guard lemma gives `P_ba,ba <= I/1600` at every regular boundary. Here `bhat_a=0`, so `e_ba=b_a` forever, and

```
V >= 1600 ||b_a||^2
  = 38468153689/625062500
  = 61.54289161323867... > 36.
```

Consequently no eventual capture-and-retention in V<=36 exists for this admitted history once the regular covariance bound applies. The actual projection remains inactive because the estimate is zero. Failure of its sufficient storage guard does not mean projection is active.

More generally the same small-rotation family allows physical bias magnitude approaching B_a. A universal eventual storage-ball radius must therefore permit `sqrt(V)>=40 B_a`, about 9.0066642. The current sufficient projection-inactivity guard only covers radii below `40(.4-B_a)`, about 6.9933358. These cannot overlap because `2 B_a>.4`. Thus universal capture into a ball covered by THIS particular sufficient guard is not a viable completion route under the current assumptions.

This is NOT a refutation of six-degree TILT capture (the witness tilt is only 1.15 degrees), practical stability, nominal force/field separation or B_*. Nominal force and field are perpendicular and nominal gyro bias is zero. Keep the small guard as a correct local result, but do not equate it with a universally capturable practical region; the larger/shaped region or literal projection sector must be handled within the same proof path, without weakening physical assumptions or quality gates.

The physical bias does not obey the estimator OU law: its exact error prediction is `phi*b_a+(1-phi)*b_a=b_a`. Omitting the existing `(1-phi)b_a` mismatch would incorrectly hide this example.

### Validation and remaining status

The standalone exact-arithmetic checks now total 25 passing tests: 13 signed-integral/physical-weight tests, 5 adjoint-compatibility tests, and 7 quiet-bias/service/guard tests. They are NOT a repository-wide CI or native replay result. No all-time float32/integer-clock/arithmetic-totality claim is made. No repository commit or theorem flag was changed by this continuation.

The construction-linked signed nominal margins, common historical B_*, full covariance upper bound, J, useful rho0, and complete nonlinear retained-region/capture theorem remain open. The new guard obstruction prevents one downstream dead end; it does not substitute for closing the upstream same-history dependency.

mgrouch:
Updated PR #561 onto current main with merge commit 89574e5f, preserving the proof changes while taking current main as the merge parent. The failure in run 35541814741 is Ruff B023 in tests/validation/test_ou3_article_figure_layout.py: the savefig side-effect lambda closes over the loop-local `figures`. Fixed by binding that list as a lambda default; no test semantics or quality gate changed.

Proof continuation at ad5b9962: the signed temporal route now records the exact adjoint compatibility needed to eliminate the innovation functional: Z_i=Z_(i+1)A_i and W_i=Z_(i+1)K_i on the same chronological execution. It also retains the actual v/p/S correction jumps in the weighted nominal integral identity. Separately, an admitted quiet constant-bias history proves V<=36 cannot be a universal capture target: with R=Rx(2 atan(1/100)) and b_a=g(R^T e_z-e_z), |b_a|≈.1961232 is within the existing physical bound, the measured level/north record is unchanged, and the established P_ba,ba<=I/1600 gives V>=1600|b_a|^2>61.54289. This does not refute six-degree tilt capture or practical stability; it means the local projection-inactive guard must not be promoted to the universal retained region. No assumptions or shipping behavior changed.

Uniform signed separation, the gyro first-exit barrier, common B_*, J, useful rho0 and nonlinear capture/retention remain open; finite 400–600 s audits are still diagnostic only.

